PR #26, opened 2023-09-08
github actions for releases and code scanning

anmarchenko:
**What does this PR do?**
preparing for releases support and adding relevant GitHub actions

**Motivation**
Release 0.1.0 incoming

**Additional notes**
See my question on pushing gem to github packages - when does it happen?

**How to test the change?**
Did I miss anything?


.github/workflows/build-gem.yml
```diff
@@ -0,0 +1,165 @@
+name: Build gem
+
+on:
+  workflow_dispatch:
+    inputs:
+      push:
+        description: Push gem
```
anmarchenko, line 7: where does this input come from? under what conditions the gem is pushed to github packages?

TonyCTHsu: `workflow_dispatch` means that it can be manually triggered from Github UI

.github/workflows/build-gem.yml
```diff
@@ -0,0 +1,165 @@
...
+        required: true
+        type: boolean
+        default: true
+  push:
+    branches:
+      - '**'
```
TonyCTHsu, line 13: Is it necessary to build the gem on every push on any branch? Wonder what's would be your thoughts on this?

TonyCTHsu (approved):

PR #28, opened 2023-09-12
improve contribution documentation

anmarchenko:
**What does this PR do?**
- Documents test matrix
- Documents how to work with appraisals

**Motivation**
Improving edveloper experience

### DIFF
--- a/docs/DevelopmentGuide.md
+++ b/docs/DevelopmentGuide.md
@@ -115,8 +115,8 @@ To install dependencies, run `bundle exec appraisal install`.
 In addition, if you already know which appraisal group definition to work with, you can target a specific group operation with environment vairable `APPRAISAL_GROUP`, instead of all the groups from your environment. For example:
 
 ```bash
-# This would only install dependencies for `rspec` group definition
-APPRAISAL_GROUP=rspec bundle exec appraisal install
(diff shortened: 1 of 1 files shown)


docs/DevelopmentGuide.md
```diff
@@ -69,14 +92,33 @@ bundle exec rake test:main
...
+
+After introducing a new group definition or changing existing one, run `bundle exec appraisal generate` to propagate the changes.
+
+To install dependencies, run `bundle exec appraisal install`.
+
+In addition, if you already know which appraisal group definition to work with, you can target a specific group operation with environment vairable `APPRAISAL_GROUP`, instead of all the groups from your environment. For example:
+
+```bash
+# This would only install dependencies for `rspec` group definition
+APPRAISAL_GROUP=rspec bundle exec appraisal install
```
TonyCTHsu, line 119: I think `rspec` is not a good example, because it is declared from the `Gemfile` at the project root. 
Perhaps using `cucumber` make more sense

TonyCTHsu (approved):

docs/DevelopmentGuide.md
anmarchenko: true, will replace with cucumber

PR #40, opened 2023-10-02
[CIAPP-4278] Fix an issue with emojis in commit message breaking LocalGit tags provider

anmarchenko:
**What does this PR do?**
Fixes an issue with emojis in git commit message.

Sometimes Ruby sets default external encoding to US_ASCII even thou LANG is "en_US.UTF-8". In these cases non-ASCII characters like emojis break LocalGit provider.

In order to mitigate that, I force UTF-8 encoding if external encoding is US-ASCII. It is safe to do as UTF-8 is backwards compatible with US-ASCII.

**Motivation**
Fixing bug report

**How to test the change?**
By running unit tests


ekump (approved):

ivoanjo:
Thanks for fixing this so I can litter commit messages with emoji again :tada:

PR #51, opened 2023-10-16
[CIVIS-2467] Support test visibility protocol via Datadog Agent with EVP proxy

anmarchenko:
**What does this PR do?**
For the case when agentless mode is not configured, we check whether Datadog Agent has EVP (event platform) proxy endpoint and if it does, we use test visibility protocol via this proxy to submit traces to Datadog.

This PR introduces Datadog::CI::Transport::Api module that incapsulates API settings such as host, port, ssl usage, gzip compression, etc.

Datadog::CI::TestVisibility::Transport no longer constructs API layer directly: api gets injected in the transport on Components initialization phase.

**Motivation**
Implementing the next feature for CI visibility library on a road to GA.

**How to test the change?**
Run ruby tests locally without DD_CIVISIBLITY_AGENTLESS_ENABLED

### DIFF
--- a/lib/datadog/ci/transport/api/base.rb
+++ b/lib/datadog/ci/transport/api/base.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Datadog
+  module CI
+    module Transport
+      module Api
+        class Base
+          def request(path:, payload:, verb: "post")
+          end
+
+          private
+
+          def headers
+            @headers ||= {
+              Ext::Transport::HEADER_CONTENT_TYPE => Ext::Transport::CONTENT_TYPE_MESSAGEPACK
+            }
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/datadog/ci/transport/api/base.rb
+++ b/lib/datadog/ci/transport/api/base.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "../../ext/transport"
+
 module Datadog
   module CI
     module Transport
--- a/lib/datadog/ci/transport/api/base.rb
+++ b/lib/datadog/ci/transport/api/base.rb
@@ -11,7 +11,7 @@ module Datadog
           private
 
           def headers
-            @headers ||= {
+            {
               Ext::Transport::HEADER_CONTENT_TYPE => Ext::Transport::CONTENT_TYPE_MESSAGEPACK
             }
           end
--- a/lib/datadog/ci/transport/api/base.rb
+++ b/lib/datadog/ci/transport/api/base.rb
@@ -5,7 +5,15 @@ module Datadog
     module Transport
       module Api
         class Base
+          attr_reader :http
+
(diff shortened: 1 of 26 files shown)


lib/datadog/ci/transport/api/builder.rb
```diff
@@ -0,0 +1,31 @@
...
+require_relative "ci_intake"
+require_relative "evp_proxy"
+
+module Datadog
+  module CI
+    module Transport
+      module Api
+        module Builder
+          def self.build_ci_test_cycle_api(settings)
+            dd_site = settings.site || "datadoghq.com"
```
romainkomorndatadog, line 12: This could (or should?) come from constants (like the hot prefix below?)

romainkomorndatadog (approved):

lib/datadog/ci/transport/api/builder.rb
anmarchenko: this is true, will define a DEFAULT_DD_SITE constant

marcotc (approved):
The code looks good.

Given this is a sizeable new feature, would it make sense if we had at least one integration test for it? (unless there's already one and I interpreted it as unit test)

anmarchenko:
@marcotc we run integration tests in https://github.com/DataDog/test-environment but this is a good point, I kind of neglected integration tests that we have in this repo

I will add a task for myself to add more of them

marcotc:
@anmarchenko if this feature is covered by https://github.com/DataDog/test-environment, then all good!

lib/datadog/ci/transport/api/ci_intake.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require_relative "base"
+require_relative "../http"
+
+module Datadog
+  module CI
+    module Transport
+      module Api
+        class CIIntake < Base
```
marcotc, line 10: `CIIntake` 😛

anmarchenko: Looks ugly, I know :( EVPProxy looks not so much better
I think I should drop abbreviations and call them CiIntake and EvpProxy

lib/datadog/ci/transport/api/base.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Datadog
+  module CI
+    module Transport
+      module Api
+        class Base
+          def request(path:, payload:, verb: "post")
+          end
```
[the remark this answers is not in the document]
anmarchenko: thanks, useful suggestion again @GustavoCaso, this can further decouple APIs from underlying transport used, will try to refactor it today

lib/datadog/ci/transport/api/ci_intake.rb
anmarchenko: The official name of the intake is CiTestCycle, I will use it

PR #56, opened 2023-10-23
upgrade github-script action to v6

### LINKED ISSUE #35: Draft release action - deprecation message
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ - draft release action

anmarchenko:
**What does this PR do?**
Closes #35

Update github-script action to a newer version to fix deprecation message

**Motivation**
Deprecation message appears every time this action is used

**How to test**
https://github.com/DataDog/datadog-ci-rb/actions/runs/6614787863/job/17965504994
there is no deprecation message for the latest run anymore

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -8,14 +8,14 @@ jobs:
     steps:
       - name: Get release version
         id: releaseVersion
-        uses: actions/github-script@7a5c598405937d486b0331594b5da2b14db670da # 6.1.0
+        uses: actions/github-script@v6
         with:
           result-encoding: string
           script: |
             return '${{ github.event.ref }}'.substring('bump_to_version_'.length);
       - name: Get milestone for version
         id: milestone
-        uses: actions/github-script@7a5c598405937d486b0331594b5da2b14db670da # 6.1.0
+        uses: actions/github-script@v6
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/draft-release.yml
```diff
@@ -8,14 +8,14 @@ jobs:
     steps:
       - name: Get release version
         id: releaseVersion
-        uses: actions/github-script@7a5c598405937d486b0331594b5da2b14db670da # 6.1.0
+        uses: actions/github-script@v6
```
ivoanjo, line 11: Should we pin to the latest version? I believe the intention of the previous approach was exactly to not automatically update to whatever the action author uploads, but instead to keep on a specific version until we decide to move.

marcotc: If github has been a good citizen regarding major versions on their GH actions, I don't see why not trust their semver.

anmarchenko: I volunteer for this project to be a guinea pig and replace all git-sha with versions - if it breaks, this is not going to be a major issue, this library is maintained by a single team/person

romainkomorndatadog (approved):

PR #61, opened 2023-10-30
Manual test tracing API

anmarchenko:
**What does this PR do?**
This PR refactors `Datadog::CI` internals and provides a public API for manual test instrumentation via `Datadog::CI` module.

How to use the new API:

```ruby
# trace a test
Datadog::CI.trace_test(
  "my test",
  tags: {
    "test.framework" => "my framework",
    "test.suite" => "my suite"
  },
  service_name: "service",
  operation_name: "test.test"
) do |test|
   # test here
   
   if result == :pass
     test.passed!
   else
     test.failed!
   end
end

# trace any custom span inside the test
Datadog::CI.trace(
  "step",           # span type
  "my step",     # span name
  tags: {
      "test.framework" => "my framework",
  }
) do |span|
  # code here
end
```

There is also a convenience method `CI.start_test` that does not accept a block:

```ruby
my_test = Datadog::CI.start_test(
  "my test",
  tags: {
    "test.framework" => "my framework",
    "test.suite" => "my suite"
  },
  service_name: "service",
  operation_name: "test.test"
)

# ....

my_test.finish
```

These methods return or yield new `Datadog::CI::Span` object that represents tracing span in CI context adding some domain-specific behaviour (like a possibility to set status with `passed!`, `skipped!`, and `failed!` methods)

**Next steps (in subsequent PRs)**
- create `Datadog::CI::ContextProvider` module to store `active_test` span
- expand manual API to support TestSession, TestModule, and TestSuite traces
- create domain objects for Test, TestSuite, etc (like we have for Span now)
- use the manual API to instrument RSpec, Minitest, Cucuber with test suite level visibility
- serialize and submit test suite level visibility spans to the backend

**How to test the change?**
No behaviour change, unit&integration tests make sure that it still works


lib/datadog/ci/contrib/minitest/hooks.rb
```diff
@@ -19,41 +18,37 @@ def before_setup
...
+                CI::Ext::Test::TAG_FRAMEWORK_VERSION => CI::Contrib::Minitest::Integration.version.to_s,
+                CI::Ext::Test::TAG_TYPE => Ext::TEST_TYPE,
+                CI::Ext::Test::TAG_SUITE => test_suite
+              },
+              service_name: configuration[:service_name],
+              operation_name: configuration[:operation_name]
             )
 
-            Thread.current[:_datadog_test_span] = span
+            Thread.current[:_datadog_test_span] = test_span
```
marcotc, line 33: A bit of an unrelated comment, but this `before_setup` method cannot be nested with itself right? I ask because setting `Thread.current[:_datadog_test_span] = test_span` twice before calling the respective `after_teardown` in the correct order will cause issues.

marcotc (approved):

lib/datadog/ci/contrib/minitest/hooks.rb
anmarchenko: good point, I will check that - should not be the case, I have a bunch of minitest projects for integration testing, did not encounter lost test traces

anmarchenko: anyway this is something to replace with `Datadog::CI::Context` in the subsequent PR

PR #64, opened 2023-11-03
[CIVIS-2903] Add fiber-local context and public API to get active test

anmarchenko:
**What does this PR do?**
Adds public `Datadog::CI.active_test` and `Datadog::CI.active_span` methods in order to provide an official way to get currently running span or test.

This will allow us to get rid of adhoc saving the current test in our instrumentations, for example like this:

```ruby
# cucumber
@current_feature_span = Datadog::CI.start_test()
# ...
@current_feature_span.finish
```

or like this 

```ruby
# minitest
test_span = Datadog::CI.start_test()
Thread.current[:_datadog_test_span] = test_span
# ...
test_span = Thread.current[:_datadog_test_span]
test_span.finish
```

Now it should be done like this:
```ruby
Datadog::CI.start_test()
# ...
test_span = Datadog::CI.active_test
test_span.finish
```

`active_test` is stored in CI-specific fiber-local context (see `Datadog::CI::Context::Local`)
`active_span` is fetched using existing `Datadog::Tracing.active_span` method

**Motivation**
Finishing the manual API for the current state of test-level visibility, start to handle CI-specific context in this library

**How to test the change?**
By running instrumented Ruby project and making sure nothing has changed in submitted tests


lib/datadog/ci/context/local.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        def self.next_instance_id
+          UNIQUE_INSTANCE_MUTEX.synchronize { UNIQUE_INSTANCE_GENERATOR.next }
+        end
+
+        private
+
+        def active_test=(test)
+          Thread.current[@key] = test
+        end
```
ivoanjo, line 55: I wonder if we could get away with not having a unique id per context. Is it really that common that we'll have more than one context alive in an app?

I'm asking because this has the same potential memory leak issues I shared in our private chat; namely, that the unique keys will live forever until the Ruby process dies.

anmarchenko: it is very convenient for testing though, not sure how often would it be used in real world case

what is important for us is that CI session is short lived so there would be one recorder instance per forked process and there is no real danger in creating really bad memory leak in my opinion

ivoanjo: Note that I'm not saying to create a single context forever; what I'm saying is, if the semantics is that in the usual case only one context would be alive, it seems like you could reuse the key?

anmarchenko: yes, this is true, let me try to do that and see how many tests break...

anmarchenko: @ivoanjo it worked without issues and made the code simpler! thank you for the suggestion

romainkomorndatadog (approved):

lib/datadog/ci/recorder.rb
```diff
@@ -40,30 +62,59 @@ def trace(span_type, span_name, tags: {}, &block)
           span_type: span_type
         }
 
-        create_datadog_span(span_name, span_options: span_options, tags: tags, &block)
-      end
-
-      private
-
-      def create_datadog_span(span_name, span_options: {}, tags: {}, &block)
+        # create_datadog_span(span_name, span_options: span_options, tags: tags, &block)
```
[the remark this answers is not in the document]
anmarchenko: thank you! this should not be here

README.md
```diff
@@ -30,18 +30,18 @@ To activate `RSpec` integration, add this to the `spec_helper.rb` file:
...
-  # Configures the tracer to ensure results delivery
-  c.ci.enabled = true
-
-  # The name of the service or library under test
-  c.service = 'my-ruby-app'
-
-  # Enables the RSpec instrumentation
-  c.ci.instrument :rspec, **options
+# Only activates test instrumentation on CI
+if ENV["DD_ENV"] == "ci"
```
marcotc, line 34: Offtopic: what do you think of making this check `ENV["DD_ENV"] == "ci"` internally, and assume most clients want this? And then provide an option to override it, for the exceptional clients that don't want this.

romainkomorndatadog: I don't think making internal assumptions on when to turn things on or off based on `DD_ENV` would make sense to me as a user. It's much more likely I'd end up wondering why things aren't working because I missed the part in the docs that says `DD_ENV` has to be set and equal to `ci` for things to work.

anmarchenko: agree with @romainkomorndatadog, there is a very fine line between doing "magic" and surprising customers with unexpected default behaviour. Many clients use "test" as a DD_ENV value for CI for example. Here in the docs we just want to point out that they should enable CI visibility mode only when running on CI to avoid local test runs to be reported.

PR #65, opened 2023-11-06
Remove _dd.measured tag from spans

anmarchenko:
**What does this PR do?**
Removes "_dd.measured" tags from spans in CI visibility tracer.

**Motivation**
_dd.measured tag is used for automatically generating metrics out of non-root traces. This feature is not used for CI visbility, other test tracers do not send this tag so it can be removed

**How to test the change?**
No visible change in behaviour

### DIFF
--- a/lib/datadog/ci/recorder.rb
+++ b/lib/datadog/ci/recorder.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "datadog/tracing"
-require "datadog/tracing/contrib/analytics"
 
 require "rbconfig"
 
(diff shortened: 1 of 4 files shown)


lib/datadog/ci/recorder.rb
```diff
@@ -48,23 +47,22 @@ def trace(span_type, span_name, tags: {}, &block)
 
       def create_datadog_span(span_name, span_options: {}, tags: {}, &block)
         if block
-          ::Datadog::Tracing.trace(span_name, **span_options) do |tracer_span, trace|
+          Datadog::Tracing.trace(span_name, **span_options) do |tracer_span, trace|
```
romainkomorndatadog, line 50: Did I googled correctly if my understanding is that this "moves" resolution for `Datadog` from global to local?

Did the analytics class need to be global before, and since you're not using it anymore in this context, it's now safe to use the imported reference? Or was it already unnecessary to begin with?

romainkomorndatadog (approved):
Added a non-blocking question, but other than that, LGTM.

lib/datadog/ci/recorder.rb
anmarchenko: it was unnecessary from the beginning and tbh I am just bad with this kind of stuff and very often use `::` prefix where I don't have to

PR #75, opened 2023-11-29
Skip tracing when CI mode disabled and manual API is used

anmarchenko:
**What does this PR do?**
Couple of small but important fixes regarding CI mode configuration:
- currently when using manual API and setting `settings.ci.enabled` to `false` (when running tests locally) leads to a complete failure because `@ci_recorder` is nil. Now we are creating the recorder in disabled state in this case to stop tracing but preserving the behaviour. 
  - In order to make it more bearable without adding null checks everywhere a new span type is provided - `NullSpan` which is an implementation of NullObject pattern.
- we will disable test suite level visibility always when we detect old agent version that does not support event protocol

**How to test the change?**
Unit tests are provided

### DIFF
--- a/lib/datadog/ci.rb
+++ b/lib/datadog/ci.rb
@@ -29,7 +29,8 @@ module Datadog
       # @param [String] service_name the service name for this session
       # @param [Hash<String,String>] tags extra tags which should be added to the test.
       # @return [Datadog::CI::TestSession] returns the active, running {Datadog::CI::TestSession}.
-      # @return [nil] if test suite level visibility is disabled (old Datadog agent detected)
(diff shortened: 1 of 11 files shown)


romainkomorndatadog (approved):

romainkomorndatadog:
Are null spans dropped by the tracer's "writer" (whatever the component that actually sends span data to our intakes is called in the Ruby tracer)?

anmarchenko:
> Are null spans dropped by the tracer's "writer" (whatever the component that actually sends span data to our intakes is called in the Ruby tracer)?

Datadog's internal Tracer does not know anything about them, they are just created in place to avoid NullPointerException, so they are not kept in any context and are not submitted anywhere

PR #87, opened 2023-12-11
do not collect environment tags when CI is not enabled

anmarchenko:
Fixes https://github.com/DataDog/dd-trace-rb/issues/3314

**What does this PR do?**
Prevents Datadog::CI::Ext::Environment from collecting environment information and validating it if CI mode is not enabled.

**Motivation**
Regression introduced in 0.5.0 release: ddtrace started to emit error logs in production about CI environment tags not being provided

**How to test the change?**
Run ruby application with ddtrace in production mode

### DIFF
--- a/lib/datadog/ci/test_visibility/recorder.rb
+++ b/lib/datadog/ci/test_visibility/recorder.rb
@@ -31,7 +31,7 @@ module Datadog
           @enabled = enabled
           @test_suite_level_visibility_enabled = enabled && test_suite_level_visibility_enabled
 
-          @environment_tags = Ext::Environment.tags(ENV).freeze
+          @environment_tags = @enabled ? Ext::Environment.tags(ENV).freeze : {}
           @local_context = Context::Local.new
           @global_context = Context::Global.new
         end
--- a/spec/datadog/ci/test_visibility/recorder_spec.rb
+++ b/spec/datadog/ci/test_visibility/recorder_spec.rb
@@ -50,6 +50,21 @@ RSpec.describe Datadog::CI::TestVisibility::Recorder do
       let(:ci_enabled) { false }
     end
 
+    describe "#initialize" do
+      subject do
+        described_class.new(
+          enabled: ci_enabled,
+          test_suite_level_visibility_enabled: experimental_test_suite_level_visibility_enabled
+        )
+      end
+
+      it "doesn't collect environment tags" do
+        expect(Datadog::CI::Ext::Environment).not_to receive(:tags)
+
+        subject
+      end
+    end
+
     describe "#trace_test" do
       context "when given a block" do
         let(:spy_under_test) { spy("spy") }


lib/datadog/ci/test_visibility/recorder.rb
```diff
@@ -31,7 +31,7 @@ def initialize(enabled: true, test_suite_level_visibility_enabled: false)
           @enabled = enabled
           @test_suite_level_visibility_enabled = enabled && test_suite_level_visibility_enabled
 
-          @environment_tags = Ext::Environment.tags(ENV).freeze
+          @environment_tags = @enabled ? Ext::Environment.tags(ENV).freeze : {}
           @local_context = Context::Local.new
           @global_context = Context::Global.new
         end
```
ivoanjo, line 37: I haven't looked at the CI stuff in a while, so pardon me if I'm asking a stupid question but... I don't understand -- why is a `TestVisibility::Recorder` being instanced at all for folks which are not running CI?

E.g. have you considered doing something like what we're doing for profiling

https://github.com/DataDog/dd-trace-rb/blob/22c03703a37362cfc930388f659cda4fde8fc60e/lib/datadog/profiling/component.rb#L15

...where we don't instance anything if it's off?

I'm asking because this seems to have a potential to be a sharp edge. E.g. if we're skipping things one-by-one when CI is disabled, it seems easy to "oops" other things that shouldn't be loaded as well, and it seems a bit of a potential whack-a-mole of needing to remember "hey this gets instanced even when CI is not in use, so remember to disable the things that need to be disabled!".

nikita-tkachenko-datadog (approved):

lib/datadog/ci/test_visibility/recorder.rb
anmarchenko: yes, I think this was not the right call
let us ship this fix now and I made a note to remove Recorder instance completely when CI was not enabled

My "genious" thinking was that Datadog::CI calls will fail when there is no recorder instance when running tests with CI instrumentation disabled. This is an expected use case: we enable CI instrumentation only in CI environments and we run tests locally without instrumentation. So this means that when users use Datadog::CI entrypoint to manually instrument their tests we need to make sure that these calls do not fail. As I hate nil checks everywhere, I decided to just let Recorder be there, just skip traces creation when CI is not enabled. 

Of course, I did not remember the fact that Recorder's constructor has a side effect: it collects environment tags and validates them. It is done in the constructor for performance reasons to make sure that they are being collected in the beginning of test run and cached for the whole run with no risk of this call being made multiple times.

I guess better way to make Datadog::CI calls work when CI instrumentation is disabled would be to introduce ni-checks or introduce a NullRecorder class that would return NullSpan all the time to skip tracing. Anyway, this is a topic for a follow-up PR and I will definitely do it before 0.6.0 version of datadog-ci gem.

ivoanjo (approved):
Ack, seems reasonable, here sir have my approval :+1:

PR #92, opened 2023-12-15
[CIVIS-7948] Test suite level visibility instrumentation for Minitest framework

anmarchenko:
**What does this PR do?**
Adds test suite level visibility instrumentation for Minitest framework

**Important design note**
The most important decision made during the development was about the concurrent test execution in Minitest. 

When tests are executed serially, they are executed suite by suite (or Runnable by Runnable in Minitest case) just like in RSpec or Cucumber. Things change when tests are executed concurrently (which is a built in feature for Minitest). In this case runners (implemented as threads or forked processes) receive jobs and execute them one by one. "Job" in this case is a single test (or Runnable's method) not the whole runnable:
https://github.com/minitest/minitest/blob/master/lib/minitest/parallel.rb#L33

The proposed implementation is a pragmatic solution to this problem:
- for serial execution we treat Runnable as a test suite that gives us the same functionality as for RSpec and Cucumber frameworks
- for parallel execution we wrap every test in a separate test suite (adding a unique prefix "(#{test_name} concurrently)")

This produces the following session traces

For serial execution:
image

For concurrent execution:
image

This approach does not allow to track suite performance in concurrent mode but it shows the execution model better: 
with concurrent execution enabled the most important performance metric is time to execute individual tests, not test files.

**Other Notes**
- Using DatadogReporter for test suites tracking isn't possible due to the fact that Rails parallel executor's workers do not call reporter's callbacks: https://github.com/rails/rails/blob/f481353150c100dbd944d5aa84f2480264cd723c/activesupport/lib/active_support/testing/parallelization/worker.rb#L50
- During testing I hit the limitation of tracer's buffer size: concurrent threads running a lot of tests overflow the buffer, and traces are being dropped in result. This PR increases buffer size for CI mode to 10 000 traces.

**How to test the change?**
Tested using the following repositories:
https://github.com/anmarchenko/openstreetmap-website
https://github.com/anmarchenko/sidekiq
https://github.com/anmarchenko/feedbin
https://github.com/anmarchenko/devdocs
https://github.com/anmarchenko/middleman


lib/datadog/ci/contrib/minitest/runnable.rb
```diff
@@ -0,0 +1,46 @@
...
+              return super unless datadog_configuration[:enabled]
+              return super if parallel?
+
+              method = runnable_methods.first
+              return super if method.nil?
+
+              test_suite_name = Suite.name(self, method)
+
+              test_suite = Datadog::CI.start_test_suite(test_suite_name)
+              test_suite.passed! # will be overridden if any test fails
```
nikita-tkachenko-datadog, line 23: Does it mean a test suite can never have 'skipped' status? Not sure what the intended behaviour is, mentioning it just in case

lib/datadog/ci/contrib/minitest/suite.rb
```diff
@@ -0,0 +1,20 @@
...
+module Datadog
+  module CI
+    module Contrib
+      module Minitest
+        # Minitest integration constants
+        # TODO: mark as `@public_api` when GA, to protect from resource and tag name changes.
+        module Suite
+          def self.name(klass, method_name)
+            source_location, = klass.instance_method(method_name).source_location
+            source_file_path = Pathname.new(source_location.to_s).relative_path_from(Pathname.pwd).to_s
```
nikita-tkachenko-datadog, line 12: Looks like it's implicitly assumed here that the current dir is always the same as the repo root. Not sure if the two can be different in Ruby world. If they can, things like source code integration and ITR will break in the backend

nikita-tkachenko-datadog (approved):

lib/datadog/ci/contrib/minitest/runnable.rb
anmarchenko: Yes, I haven't implemented `SKIP` for test suites in any of the instrumentations right now. Is it heavily used by other libraries (if we don't include ITR)?

lib/datadog/ci/contrib/minitest/suite.rb
anmarchenko: I haven't yet encountered cases where this wouldn't be true for Ruby as both `bin/rails` and `rake` command are always executed from the root. Note that this is actually a test suite name (which could be any arbitrary string), not the source code integration (which I will add together with ITR). For source code integration, I will check for the repository root.

lib/datadog/ci/contrib/minitest/runnable.rb
nikita-tkachenko-datadog: In Java the suite status is SKIP if it contains 0 test cases, or if every test case in it is skipped. Not sure about other languages

anmarchenko: thank you! I will add a note for the future improvements

PR #97, opened 2024-01-05
add note to README.md about test suite level visibility feature

anmarchenko:
**What does this PR do?**
Added a note about how to configure experimental test suite level visibility feature to gather early feedback

**Motivation**
Make this feature more visibile and gather early feedback.


README.md
```diff
@@ -134,6 +134,30 @@ Datadog.configure { |c| c.ci.agentless_mode_enabled = true }
 
 ## Additional configuration
 
+### Experimental test suite level visibility support
+
+Test suite level visibility (TSLV) is available for `datadog-ci`` gem versions 0.6 and above.
```
juan-fernandez, line 139: is that a typo?
```suggestion
Test suite level visibility (TSLV) is available for `datadog-ci` gem versions 0.6 and above.
```

juan-fernandez (approved):

README.md
anmarchenko: yes, it is, thank you

ivoanjo, line 139: Minor: I find it somewhat weird to list the minimum version for a feature in the readme, since, well, every feature that gets added is a feature that has a minimum version.

anmarchenko: this is very true, I think this info is not needed here

PR #100, opened 2024-01-10
[CIVIS-7722] Parametrized tests support for cucumber

anmarchenko:
**What does this PR do?**
Adds parametrized test support for cucumber-rby (version 4 and above):
image

**Motivation**
Currently all test runs with [scenario outline](https://cucumber.io/docs/gherkin/reference/#scenario-outline) have the same test fingerprint and treated as a single test being run multiple times. We should use `test.parameters` tag to distinguish these tests from each other

**Additional Notes**
- Cucumber 3.0 does not have a way to extract examples but in Cucumber 3.0 all scenario outline runs have "(Example #N)" prefix so they are unique
- JSON hash is set using JSON.stringify because Cucumebr examples are built using strings, so I am sure that these parameters have proper string representations

**How to test the change?**
Tested using https://github.com/anmarchenko/jekyll

### DIFF
--- a/spec/datadog/ci/contrib/cucumber/instrumentation_spec.rb
+++ b/spec/datadog/ci/contrib/cucumber/instrumentation_spec.rb
@@ -216,7 +216,7 @@ RSpec.describe "Cucumber formatter" do
           expect(span.get_tag(Datadog::CI::Ext::Test::TAG_NAME)).to eq("scenario with examples")
 
           expect(span.get_tag(Datadog::CI::Ext::Test::TAG_PARAMETERS)).to eq(
-            "{\"num1\":\"#{index}\",\"num2\":\"#{index + 1}\",\"total\":\"#{index + index + 1}\"}"
+            "{\"arguments\":{\"num1\":\"#{index}\",\"num2\":\"#{index + 1}\",\"total\":\"#{index + index + 1}\"},\"metadata\":{}}"
           )
         else
           expect(span.get_tag(Datadog::CI::Ext::Test::TAG_NAME)).to eq(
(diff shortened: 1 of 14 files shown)


spec/datadog/ci/contrib/cucumber/instrumentation_spec.rb
```diff
@@ -202,14 +203,45 @@
...
+      expect(test_spans.count).to eq(3)
+      expect(test_suite_spans.count).to eq(1)
+
+      test_spans.each_with_index do |span, index|
+        # test parameters are available since cucumber 4
+        if cucumber_4_or_above
+          expect(span.get_tag(Datadog::CI::Ext::Test::TAG_NAME)).to eq("scenario with examples")
+
+          expect(span.get_tag(Datadog::CI::Ext::Test::TAG_PARAMETERS)).to eq(
+            "{\"num1\":\"#{index}\",\"num2\":\"#{index + 1}\",\"total\":\"#{index + index + 1}\"}"
```
juan-fernandez, line 219: does the UI render these parameters correctly? The format the other tracers follow is 
```
{
  "arguments": // ... your parameters
  "metadata": { // ... extra metadata }
}
```
Screenshot 2024-01-11 at 13 03 33

anmarchenko: Working on it, UI is handling it correctly:

image

anmarchenko: @juan-fernandez fixed:
image

juan-fernandez (approved):

PR #104, opened 2024-01-15
add .repository_name to Utils::Git

anmarchenko:
**What does this PR do?**
In order to simplify onboarding for new customers it provides better default service name when none is provided.

**Additional Notes**
- last segment of git remote URL is used as a repository name
- as a fallback, current folder is used (where git repo is located)

**How to test the change?**
Unit tests are provided.


lib/datadog/ci/contrib/cucumber/configuration/settings.rb
```diff
@@ -21,7 +22,9 @@ class Settings < Datadog::CI::Contrib::Settings
 
             option :service_name do |o|
               o.type :string
-              o.default { Datadog.configuration.service_without_fallback || Ext::DEFAULT_SERVICE_NAME }
+              o.default do
+                Datadog.configuration.service_without_fallback || Utils::Git.repository_name || Ext::DEFAULT_SERVICE_NAME
```
romainkomorndatadog, line 26: Should this come from a function provided by our CI gem's utils, instead of a per-framework "implementation"?

eg: `Utils::Something.get_service_name()`?

romainkomorndatadog (approved):

lib/datadog/ci/contrib/cucumber/configuration/settings.rb
anmarchenko: yes, I agree, the logic to look into Datadog.configuration and Utils::Git should be extracted, will do it

romainkomorndatadog (approved):

PR #112, opened 2024-01-23
[CIVIS-8552] ci-queue support for rspec

### LINKED ISSUE #96: Track retries of specs
**Is your feature request related to a problem? Please describe.**

Spec runners like [Shopify's ci-queue](https://github.com/Shopify/ci-queue) allow you to [re-queue failed specs](https://github.com/Shopify/ci-queue?tab=readme-ov-file#what-are-requeues). While these specs are flaky, these are not failures across two separate runs, but are failures within the current CI run.

**Describe the goal of the feature**
Track and allow filtering on retries.

**Describe alternatives you've considered**
We have monkey-patched datadog-ci to provide support

Changed https://github.com/DataDog/datadog-ci-rb/blob/main/lib/datadog/ci/contrib/rspec/example.rb#L39-L46

```ruby
         case execution_result.status
         when :passed
           test_span.passed!
         when :failed
           test_span.failed!(exception: execution_result.exception)
         when nil
           # retry spes show up as run but with a nil status
           test_span.retried!(exception: execution_result.exception)
         else
           test_span.skipped!(exception: execution_result.exception) if execution_result.example_skipped?
         end

  module Datadog
    module CI
      class Span
        def retried!(exception: nil)
         
          tracer_span.set_tag(Datadog::CI::Ext::Test::TAG_STATUS, 'Retried')
          tracer_span.set_error(exception) unless exception.nil?
        end
      end
    end
  end
```

**Additional context**
In DD
Screenshot 2024-01-04 at 12 53 13 AM

I'd be happy to submit a PR.

anmarchenko:
**What does this PR do?**
Fix #96 

Adds RSpec instrumentation support when running with ci-queue test runner. It instruments RSpec::Queue::Runner that is used by ci-queue and also creates "single-test" test suite akin to Minitest parallel tests if we detect to be running in a RSpec::Core::Example instrumented by ci-queue. Also it fixes our handling of pending tests correctly marking them as skipped and adds a lot of unit tests around all of the ways we can skip tests in RSpec

**Motivation**
We have users of this library who run their rspec test suites using ci-queue


lib/datadog/ci/contrib/rspec/example.rb
```diff
@@ -49,13 +54,25 @@ def run(*)
                   case execution_result.status
                   when :passed
                     test_span.passed!
+                    test_suite_span.passed! if test_suite_span
```
romainkomorndatadog, line 57: I think you should just force `test_suite_span` to exist 😄  (and make it some sort of error condition if it doesn't).

romainkomorndatadog (approved):

lib/datadog/ci/contrib/rspec/example.rb
anmarchenko: I tried to do that and it made the inheritance hierarchy unsound (so I needed to provided Null* subclass for every class that was cumbersome or always assume that superclass is returned which breaks Liskov substitution principle, see some more context in https://github.com/DataDog/datadog-ci-rb/pull/108 - look for NullSpan explanation).

Thankfully, I will be saved from this unruly ifs when we decouple ourselves from dd-trace-rb! I will immediately migrate this project to Ruby 2.7 syntax that will make these calls just:

```ruby
test_suite_span&.passed!
```

Neat, isn't it?

PR #129, opened 2024-02-29
Remote library configuration support and ITR::Runner

anmarchenko:
**What does this PR do?**

First step for Intelligent Test Runner implementation: implement remote library settings configuration

Main changes:
- `Agentless` and `EvpProxy` api clients support multiple backend endpoints now (citestcycle-intake.dd-site for events and api.dd-site for remote configuration)
- `RemoteSettingsApi` client for fetching the remote configuration
- `Datadog::CI::ITR::Runner` class that holds ITR configuration
- Some ITR related tags for test_session_end events: test.code_coverage.enabled, test.itr.tests_skipping.enabled, test.itr.tests_skipping.type
- when ITR is enabled in local settings, `TestVisibility::Recorder` fetches remote settings and configures ITR


lib/datadog/ci/transport/remote_settings_api.rb
```diff
@@ -0,0 +1,97 @@
...
+              "id" => Datadog::Core::Environment::Identity.id,
+              "type" => Ext::Transport::DD_API_SETTINGS_TYPE,
+              "attributes" => {
+                "service" => test_session.service,
+                "env" => @dd_env,
+                "repository_url" => test_session.git_repository_url,
+                "branch" => test_session.git_branch,
+                "sha" => test_session.git_commit_sha,
+                "configurations" => {
+                  "os.platform" => test_session.os_platform,
```
juan-fernandez, line 85: do we want to include os version here?

lib/datadog/ci/transport/remote_settings_api.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        private
+
+        def payload(test_session)
+          {
+            "data" => {
+              "id" => Datadog::Core::Environment::Identity.id,
+              "type" => Ext::Transport::DD_API_SETTINGS_TYPE,
+              "attributes" => {
+                "service" => test_session.service,
```
juan-fernandez, line 79: I think including the `test_level` here is a requirement to take advantage of the `require_git` optimization for ITR (which will allow you to not always run with code coverage). Unless you're going to switch between suite and test skipping, you probably can set it to `test`

juan-fernandez (approved):

lib/datadog/ci/transport/remote_settings_api.rb
anmarchenko: same question as above: where is test_level specified?

lib/datadog/ci/transport/remote_settings_api.rb
```diff
@@ -0,0 +1,97 @@
...
+                "env" => @dd_env,
+                "repository_url" => test_session.git_repository_url,
+                "branch" => test_session.git_branch,
+                "sha" => test_session.git_commit_sha,
+                "configurations" => {
+                  "os.platform" => test_session.os_platform,
+                  "os.arch" => test_session.os_architecture,
```
anmarchenko, line 86: @juan-fernandez is "os.arch" from the example correct? in tags we use "os.architecture"

lib/datadog/ci/transport/remote_settings_api.rb
anmarchenko: will add

anmarchenko: Hah, it turned out that `os.version` tag is not present both in ruby core tracer and in test tags for CI visibility right now. I will refrain from adding it in this PR as not to change test fingerprints unnecessary and will add this one to backlog to evaluate

PR #157, opened 2024-04-11
[CIVIS-9606] reduce ITR-induced code coverage overhead for default branch

anmarchenko:
**What does this PR do?**
Implements default branch code coverage overhead optimization: when server adds with `require_git: true` to the library settings payload, we wait for git upload to be done and then request settings again. The ITR Runner is configured with the second settings payload.

The reason for this is that if default branch does not have many relevant code changes, running per test code coverage won't give us a lot of new information and can be skipped: thus, tests in default branch will run faster without per test code coverage overhead.

**How to test the change?**
Unit tests are provided

### DIFF
--- a/lib/datadog/ci/itr/runner.rb
+++ b/lib/datadog/ci/itr/runner.rb
@@ -7,6 +7,8 @@ require_relative "../ext/transport"
 
 require_relative "../git/local_repository"
 
+require_relative "../utils/parsing"
+
 require_relative "coverage/event"
 
 module Datadog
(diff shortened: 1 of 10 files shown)


romainkomorndatadog (approved):
Love how clean this all is. Very nice. :)

In the Python tracer, I have some extra bits of logging and tests for the case where `require_git` is still `true` (eg: could happen if the metadata upload doesn't finish in time or the backend misses it for some reason), which I think is a nice-to-have, but not a blocker for approval.

anmarchenko:
@romainkomorndatadog thanks for the heads up!

PR #158, opened 2024-04-12
[CIVIS-2934] skippable tests api client

anmarchenko:
**What does this PR do?**
Add skippable tests api client: will be used in subsequent PRs to request from the backend which tests can be skipped

**How to test the change?**
Unit tests only for now

### DIFF
--- a/lib/datadog/ci/itr/skippable.rb
+++ b/lib/datadog/ci/itr/skippable.rb
@@ -2,17 +2,26 @@
 
 require "json"
 
-require "datadog/core/environment/identity"
-
 require_relative "../ext/transport"
-require_relative "../utils/parsing"
+require_relative "../ext/test"
 
 module Datadog
   module CI
-    module Transport
-      # Datadog API client
-      # Calls settings endpoint to fetch library settings for given service and env
-      class RemoteSettingsApi
+    module ITR
+      class Skippable
+        class Test
+          attr_reader :name, :suite
+
+          def initialize(name:, suite:)
+            @name = name
+            @suite = suite
+          end
+
+          def ==(other)
+            name == other.name && suite == other.suite
+          end
+        end
+
         class Response
           def initialize(http_response)
             @http_response = http_response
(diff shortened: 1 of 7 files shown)


lib/datadog/ci/itr/skippable.rb
```diff
@@ -0,0 +1,113 @@
...
+        class Test
+          attr_reader :name, :suite
+
+          def initialize(name:, suite:)
+            @name = name
+            @suite = suite
+          end
+
+          def ==(other)
+            name == other.name && suite == other.suite
```
romainkomorndatadog, line 21: Are you missing the bundle (which is the module) here?

lib/datadog/ci/itr/skippable.rb
```diff
@@ -0,0 +1,113 @@
...
+          {
+            "data" => {
+              "type" => Ext::Transport::DD_API_SKIPPABLE_TESTS_TYPE,
+              "attributes" => {
+                "test_level" => Ext::Test::ITR_TEST_SKIPPING_MODE,
+                "service" => test_session.service,
+                "env" => @dd_env,
+                "repository_url" => test_session.git_repository_url,
+                "sha" => test_session.git_commit_sha,
+                "configurations" => {
```
romainkomorndatadog, line 100: Do you plan on adding support for custom configurations in a separate PR?

lib/datadog/ci/itr/skippable.rb
anmarchenko: ruby library does not use modules

anmarchenko: so I just use the same approach as JS library: only take name and suite from the payload

lib/datadog/ci/itr/skippable.rb
anmarchenko: yes, I will: https://datadoghq.atlassian.net/browse/CIVIS-9333

lib/datadog/ci/itr/skippable.rb
romainkomorndatadog: Interesting!

romainkomorndatadog (approved):

PR #171, opened 2024-05-03
[SDTEST-138] code coverage extension fixes and improvements

anmarchenko:
**What does this PR do?**
Applies the following changes and optimizations to per test code coverage:
- removes lines code coverage mode as it is not used and is not going to be used in the near future, so it is just a dead code right now
- caches last seen filename pointer so that if we cover sequential lines in the same file we skip prefix check and saving file name in the coverage hash table
- `dd_cov_data->root` and `dd_cov_data->ignored_path` are now copied (once) in malloc'd memory and stored as C-strings
- to avoid string copying, `rb_profile_frames` function is now used to store filename in resulting hash: it improves performance and solves UTF-8 filenames issue

**How to test the change?**
Tested using test-environment integration test stand: https://github.com/DataDog/test-environment/pull/283

The following changes in performance overhead were recorded:

**Rubocop**
From [109%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/504983829) to [86%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/505079404) => 27% improvement

**Middleman**
From [29,7%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/504983815) to [28,6%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/505079396) => no change

**Jekyll**
From [0,4%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/504983806) to [0,3%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/505079390) => no change

**devdocs**
From [16,4%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/504983821) to [16,9%](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/505079400) => no change

Overall result: it does not change anything for projects where code coverage overhead was low enough but provides meaningful improvement for the project where code coverage overhead was the largest.

### DIFF
--- a/ext/datadog_cov/datadog_cov.c
+++ b/ext/datadog_cov/datadog_cov.c
@@ -118,19 +118,6 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
   }
   dd_cov_data->last_filename_ptr = current_filename_ptr;
 
-  // if the current filename is not located under the root, we skip it
-  if (strncmp(dd_cov_data->root, c_filename, dd_cov_data->root_len) != 0)
-  {
-    return;
-  }
-
-  // if ignored_path is provided and the current filename is located under the ignored_path, we skip it too
-  // this is useful for ignoring bundled gems location
-  if (dd_cov_data->ignored_path_len != 0 && strncmp(dd_cov_data->ignored_path, c_filename, dd_cov_data->ignored_path_len) == 0)
-  {
-    return;
-  }
-
   int captured_frames = rb_profile_frames(
       0 /* stack starting depth */,
       PROFILE_FRAMES_BUFFER_SIZE,
@@ -144,6 +131,20 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
 
   VALUE filename = rb_profile_frame_path(dd_cov_data->frame_buffer[0]);
   if (filename == Qnil)
+  {
+    return;
+  }
+
+  char *filename_ptr = StringValuePtr(filename);
+  // if the current filename is not located under the root, we skip it
+  if (strncmp(dd_cov_data->root, filename_ptr, dd_cov_data->root_len) != 0)
+  {
+    return;
+  }
+
+  // if ignored_path is provided and the current filename is located under the ignored_path, we skip it too
+  // this is useful for ignoring bundled gems location
+  if (dd_cov_data->ignored_path_len != 0 && strncmp(dd_cov_data->ignored_path, filename_ptr, dd_cov_data->ignored_path_len) == 0)
   {
     return;
   }
--- a/ext/datadog_cov/datadog_cov.c
+++ b/ext/datadog_cov/datadog_cov.c
@@ -1,55 +1,46 @@
 #include <ruby.h>
 #include <ruby/debug.h>
 
-// constants
-#define DD_COV_TARGET_FILES 1
-#define DD_COV_TARGET_LINES 2
+const int PROFILE_FRAMES_BUFFER_SIZE = 1;
 
-static int is_prefix(VALUE prefix, const char *str)
+char *ruby_strndup(const char *str, size_t size)
 {
-  if (prefix == Qnil)
-  {
-    return 0;
-  }
+  char *dup;
 
-  const char *c_prefix = RSTRING_PTR(prefix);
-  if (c_prefix == NULL)
-  {
-    return 0;
-  }
+  dup = xmalloc(size + 1);
+  memcpy(dup, str, size);
+  dup[size] = '\0';
 
-  long prefix_len = RSTRING_LEN(prefix);
-  if (strncmp(c_prefix, str, prefix_len) == 0)
-  {
-    return 1;
-  }
-  else
-  {
-    return 0;
-  }
+  return dup;
 }
 
 // Data structure
 struct dd_cov_data
 {
-  VALUE root;
-  VALUE ignored_path;
-  int mode;
+  char *root;
+  long root_len;
+
+  char *ignored_path;
+  long ignored_path_len;
+
   VALUE coverage;
+
+  VALUE *frame_buffer;
+  uintptr_t last_filename_ptr;
 };
 
 static void dd_cov_mark(void *ptr)
 {
   struct dd_cov_data *dd_cov_data = ptr;
   rb_gc_mark_movable(dd_cov_data->coverage);
-  rb_gc_mark_movable(dd_cov_data->root);
-  rb_gc_mark_movable(dd_cov_data->ignored_path);
 }
 
 static void dd_cov_free(void *ptr)
 {
   struct dd_cov_data *dd_cov_data = ptr;
-
+  xfree(dd_cov_data->root);
+  xfree(dd_cov_data->ignored_path);
+  xfree(dd_cov_data->frame_buffer);
   xfree(dd_cov_data);
 }
 
@@ -57,8 +48,6 @@ static void dd_cov_compact(void *ptr)
 {
   struct dd_cov_data *dd_cov_data = ptr;
   dd_cov_data->coverage = rb_gc_location(dd_cov_data->coverage);
-  dd_cov_data->root = rb_gc_location(dd_cov_data->root);
-  dd_cov_data->ignored_path = rb_gc_location(dd_cov_data->ignored_path);
 }
 
 const rb_data_type_t dd_cov_data_type = {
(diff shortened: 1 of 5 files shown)


nikita-tkachenko-datadog (approved):

ivoanjo (commented):
Left some notes!

Big props for going ahead and removing unused things -- git always remembers, and less code means less bugs :)

ext/datadog_cov/datadog_cov.c
```diff
@@ -159,6 +120,12 @@ static VALUE dd_cov_start(VALUE self)
   return self;
 }
 
+static int insert_into_result(st_data_t filename, st_data_t _val, st_data_t coverage)
+{
+  rb_hash_aset((VALUE)coverage, rb_str_new2((char *)filename), Qtrue);
+  return ST_CONTINUE;
+}
```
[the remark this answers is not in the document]
anmarchenko: Thanks for pointing it out! yes, I kind of suspected that this is a bit fishy.

I will try the approach with replacing `st.h` with Ruby hash again but storing VALUE returned by `rb_profile_frames` to avoid creating new strings: it could provide me the same performance win.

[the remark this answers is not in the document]
ivoanjo: :+1: If you're going in that direction, remember to use the limit arguments in `rb_profile_frames` to get only the top entry.

ext/datadog_cov/datadog_cov.c
```diff
@@ -159,6 +120,12 @@ static VALUE dd_cov_start(VALUE self)
   return self;
```
ivoanjo, line 120: :arrow_up: Also, now that we're moving away from line coverage, maybe it would be a good idea to revisit the use of `RUBY_EVENT_LINE`? E.g. perhaps a combination of coarser-grained events (call, b_call, fiber_switch, etc) would be an overall performance win?

anmarchenko: Using RUBY_EVENT_CALL increased overhead twice: I guess this event is not optimized as good as RUBY_EVENMT_LINE

ext/datadog_cov/datadog_cov.c
```diff
@@ -1,64 +1,53 @@
 #include <ruby.h>
 #include <ruby/debug.h>
 
-// constants
-#define DD_COV_TARGET_FILES 1
-#define DD_COV_TARGET_LINES 2
+const int PROFILE_FRAMES_BUFFER_SIZE = 1;
 
-static int is_prefix(VALUE prefix, const char *str)
+char *ruby_strndup(const char *str, size_t size)
```
anmarchenko, line 6: this one is stolen from profiling extension

anmarchenko:
Found an issue with the current approach:

the code 
```ruby
class CrawlData
  delegate :last_modified, :etag, :download_fingerprint, to: :@data
```

rb_sourcefile returns `/Users/andrey.marchenko/qa/feedbin/app/models/crawl_data.rb`
rb_profile_frames returns `/Users/andrey.marchenko/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/module/delegation.rb`

so these calls are not the same and it is not safe to use one of them for checks and another one for tracking in coverage hash

ext/datadog_cov/datadog_cov.c
```diff
@@ -74,113 +63,108 @@ static VALUE dd_cov_allocate(VALUE klass)
...
   dd_cov_data->coverage = rb_hash_new();
-  dd_cov_data->root = Qnil;
-  dd_cov_data->ignored_path = Qnil;
-  dd_cov_data->mode = DD_COV_TARGET_FILES;
+  dd_cov_data->root = NULL;
+  dd_cov_data->root_len = 0;
+  dd_cov_data->ignored_path = NULL;
+  dd_cov_data->ignored_path_len = 0;
+  dd_cov_data->last_filename_ptr = 0;
+  dd_cov_data->frame_buffer = xcalloc(PROFILE_FRAMES_BUFFER_SIZE, sizeof(VALUE));
```
ivoanjo, line 73: Minor: Since we're only using a single memory position, it occurred to me a possible simplification would be to have a single `VALUE top_frame` in `dd_cov_data` and use that as a buffer instead. So this would be one less piece of dynamic memory to manage.

(This being C, you don't really need to pass in "an array" to rb_profile_frames, just a pointer to somewhere that can be written to -- and that includes a single field)

ext/datadog_cov/datadog_cov.c
```diff
@@ -74,113 +63,108 @@ static VALUE dd_cov_allocate(VALUE klass)
...
+  VALUE filename = rb_profile_frame_path(dd_cov_data->frame_buffer[0]);
+  if (filename == Qnil)
   {
     return;
   }
 
-  VALUE rb_str_source_file = rb_str_new2(filename);
-
-  if (dd_cov_data->mode == DD_COV_TARGET_FILES)
+  char *filename_ptr = StringValuePtr(filename);
```
ivoanjo, line 138: Minor: Since we're on the topic of optimization, I think you can use `RSTRING_PTR` here; `StringValuePtr` will additionally call `to_str` if its argument is not a String, but I don't think that is not relevant in this case.

ext/datadog_cov/datadog_cov.c
```diff
@@ -74,113 +63,108 @@ static VALUE dd_cov_allocate(VALUE klass)
...
     return;
   }
 
-  // this isn't optimized yet, this is a POC to show that lines coverage is possible
-  // ITR beta is going to use files coverage, we'll get back to this part when
-  // we need to implement lines coverage
-  if (dd_cov_data->mode == DD_COV_TARGET_LINES)
+  // if ignored_path is provided and the current filename is located under the ignored_path, we skip it too
+  // this is useful for ignoring bundled gems location
+  if (dd_cov_data->ignored_path_len != 0 && strncmp(dd_cov_data->ignored_path, filename_ptr, dd_cov_data->ignored_path_len) == 0)
```
ivoanjo, line 147: Minor: It occurred to me that if `ignored_path` is supposed to include `root` as its prefix, we could store only that part and compare only that part here ;)

ext/datadog_cov/datadog_cov.c
```diff
@@ -74,113 +63,108 @@ static VALUE dd_cov_allocate(VALUE klass)
...
-  // add event hook
-  rb_thread_add_event_hook(thval, dd_cov_update_line_coverage, RUBY_EVENT_LINE, self);
+  struct dd_cov_data *dd_cov_data;
+  TypedData_Get_Struct(self, struct dd_cov_data, &dd_cov_data_type, dd_cov_data);
+
+  if (dd_cov_data->root_len != 0)
+  {
+    // add event hook
+    rb_thread_add_event_hook(thval, dd_cov_update_coverage, RUBY_EVENT_LINE, self);
+  }
```
ivoanjo, line 167: Consider maybe raising an exception if root_len == 0, otherwise it can be confusing if you ever trigger this situation (especially as the codebase evolves) and then `start` silently just doesn't start which can be hard to debug.

spec/ddcov/ddcov_spec.rb
```diff
@@ -11,227 +11,201 @@ def absolute_path(path)
...
-            expect(coverage.size).to eq(1)
-            expect(coverage.keys).to include(absolute_path("calculator/calculator.rb"))
-          end
-        end
+      it "supports files with non-ASCII characters" do
+        subject.start
+        expect(I❤️Ruby.new.call).to eq("I ❤️ Ruby")
+        coverage = subject.stop
+        expect(coverage.size).to eq(1)
+        expect(coverage.keys).to include(absolute_path("calculator/code_with_❤️.rb"))
```
ivoanjo, line 54: :heart: Nice that we support this now! :D

ivoanjo (approved):
:+1: LGTM

This looks great! I've left a few final suggestions, but feel free to do them separately or not at all.

I'm assuming the performance improvements in the description are up-to-date with the move to `rb_profile_frames`, if not, I recommend doing a pass, just to make sure that some of the latest changes in the PR didn't regress the expected improvements.

anmarchenko:
@ivoanjo yes, the number for performance improvements is final and was measured with the current state of this PR. Indeed, the switch to `rb_profile_frames` caused more overhead than `rb_sourcefile` but it is safer and solves an issue with UTF-8 filenames, so I consider it being worth it.

ext/datadog_cov/datadog_cov.c
anmarchenko: Thanks, this one I will make a not of for the future improvements

PR #172, opened 2024-05-07
[CIVIS-9478] Knapsack Pro 7/RSpec 3 support

### LINKED ISSUE #147: Dropping tests with knapsack_pro
**Current behaviour**
This is a similar issue to #130 in its results, but with a different root cause (I suspect). We're observing a lot of tests and suites being discarded without a session or module ID.

We're using RSpec + [knapsack_pro](https://github.com/KnapsackPro/knapsack_pro-ruby) (in queue mode) to run tests in CI. Knapsack runs the tests in small batches which are dynamically distributed to worker nodes, which speeds up overall test runtime.

knapsack_pro 7 made [a lot of changes](https://github.com/KnapsackPro/knapsack_pro-ruby/pull/237) to the runner, which might be related. It seems like we're losing track of the session and module through this process somehow.

I've confirmed that knapsack_pro 6 is fine, so the massive PR is at fault somewhere. I've also confirmed that it only occurs in queue mode. I'm not entirely sure which project to log this on, but it feels like something that's probably easier to fix here(?)

**Expected behaviour**
Tests and suites have the correction session and module IDs and make it to Datadog.

**Steps to reproduce**
1. Setup a project with RSpec & test tracing (standard setup, nothing fancy)
2. Run tests with Knapsack Pro 7 in queue mode according to [the docs](https://docs.knapsackpro.com/knapsack_pro-ruby/guide/?rails=yes&test-runner=rspec&vcr=no&ci=github-actions#instructions), effectively setting a few env vars and running `bundle exec rake knapsack_pro:queue:rspec`

Annoyingly this requires a paid Knapsack Pro account. I'm happy to run anything to help on ours to aid in debugging.

**Environment**

* **datadog-ci-rb version:** 0.8.3
* **Configuration block (`Datadog.configure ...`):**
```ruby
require "datadog/ci"

Datadog.configure do |c|
    c.tracing.enabled = (ENV["DD_ENV"] == "ci")
    c.ci.enabled = true
    c.service = "komoju-ci"
    c.ci.instrument :rspec
end
```
* **Test framework version:** rspec 3.13.0
* **CI Provider:** GH Actions
* **Ruby version:** 3.2.3
* **Operating system:** Linux

anmarchenko:
Closes #147

**What does this PR do?**
Adds instrumentation to `KnapsackPro::Runners::Queue::RSpecRunner` for knapsack_pro gem versions 7.0 and above. This makes sure that every knapsack_pro run is correctly reported to Datadog as a separate test session.

**Motivation**
Multiple reports of Datadog dropping test spans when using knapsack_pro

**How to test the change?**
Unit tests are provided


pcothenet:
@anmarchenko Previously, we had `c.ci.force_test_level_visibility = true` to work with Knapsack Pro. Should we drop this setting when testing this Beta?

anmarchenko:
@pcothenet yes, drop force_test_level_visibility from now on - knapsack pro is supposed to be fully supported now

connorshea:
Does this also work for minitest?

anmarchenko:
@connorshea I didn't test support for minitest with knapsack pro; but it might still work. Please try it and create an issue if it doesn't!

PR #174, opened 2024-05-07
[CIVIS-9951] add settings option to ignore code coverage for bundled gems location

anmarchenko:
**What does this PR do?**
In CI environments bundled gems are often installed inside the project folder for caching purposes. Unfortunately, it makes per test code coverage for ITR extremely slow as we need to unnecessary track all the code files from the external gems.

This PR adds `DD_CIVISIBILITY_ITR_CODE_COVERAGE_EXCLUDED_BUNDLE_PATH` environment variable that allows us to set a certain path to be excluded from code coverage (supports both relative and absolute paths). Even better, for the most cases users won't have to use this configuration: we will locate bundle location and set it for them.

It works as following:
- if `bundler` gem is available in the environment, we use `Bundler.bundle_path` and set this result as ignored path
- otherwise we try to locate `vendor/bundle` or `.bundle` under the git root - if any of them exist, we ignore them

**How to test the change?**
Unit tests are provided. 

I have run test-environment with gems bundled inside the project directory to confirm the fix:
https://github.com/DataDog/test-environment/pull/284

The overhead with bundling gems in the project folder in this case is insignificantly larger: 113% vs 106% for rubocop for example. Without this change, it would be 450% vs 106%.


ext/datadog_cov/datadog_cov.c
```diff
@@ -5,10 +5,35 @@
...
+  {
+    return 0;
+  }
+}
+
 // Data structure
 struct dd_cov_data
 {
   VALUE root;
+  VALUE ignored_path;
```
ivoanjo, line 36: So I guess the same suggestion from #171 would apply here -- keeping the prefix as a `char *` and caching its length would save us some per-line work.

anmarchenko: I agree! If it's not blocking I would merge this one as is, and then optimized it together with everything else in #171

anmarchenko: I agree! If it's not blocking I would merge this one as is, and then optimized it together with everything else in #171

ivoanjo: Yeah, I don't think it's a blocker. I guess it's a slight performance regression until it's paired with #171, but I'm guessing the plan will be to release 'em together :)

anmarchenko: I will have to release it earlier than the performance improvements, but there is an important customer value in this: we have a customer that evaluates ITR and current implementation that does not ignore bundled gems installed in project folder induces up to 400% code coverage overhead.

I measured ineffectiveness of this solution using test-environment: for the worst case it increases overhead by 7% (113 vs 106) if bundled gems are installed in project folder. When gems are not installed in project folder, this change does not induce any overhead.

ivoanjo: Sounds reasonable, I'll defer to your wise judgment here :)

PR #175, opened 2024-05-13
log an error message if tracing is disabled but test visibility is enabled

anmarchenko:
**What does this PR do?**
Logs an error message when tracing is disabled but test visibility is enabled.

**Motivation**
There are many cases when tracing is disabled by competing configuration (most often in `initializers` folder to disable APM in non-production environments). Currently in this cases this library continues to work silently skipping all traces. This PR fixes that by catching attention of developer: hopefully this will lead to easier onboarding.

**How to test the change?**
Unit tests are provided

### DIFF
--- a/lib/datadog/ci/configuration/components.rb
+++ b/lib/datadog/ci/configuration/components.rb
@@ -43,6 +43,16 @@ module Datadog
         end
 
         def activate_ci!(settings)
+          unless settings.tracing.enabled
+            Datadog.logger.error(
+              "CI visibility requires tracing to be enabled. Disabling CI visibility. " \
+              "NOTE: if you didn't disable tracing intentionally, add `c.tracing enabled = true` to " \
+              "your Datadog.configure block."
+            )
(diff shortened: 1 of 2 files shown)


romainkomorndatadog (approved):

ivoanjo:
Should this perhaps be a more visible error? We have in general avoided breaking the customer's app at all costs, but if you're in CI, and have enabled CI mode, it seems kinda natural to "break" CI if something is wrong, rather than all's green and nothing works.

Maybe worth an exception (no pun intended) to our usual approach?

romainkomorndatadog:
> Maybe worth an exception (no pun intended) to our usual approach?

The "brokenness" should be "you don't get CI Visibility" , rather than "your tests don't run", so I think this is a correct approach.

The "worst case" failure for CI Vis should that it doesn't exist, so some mix of:
* your have no test results
* ITR is not working, so no tests are skipped

ivoanjo:
Fair enough, was just an idea!

PR #183, opened 2024-05-24
[SDTEST-158] Browser tests support via selenium integration

anmarchenko:
**What does this PR do?**
Automatically recognises tests that use selenium and marks them as browser tests. Integrates with Datadog RUM to automatically show browser sessions in the test run view (see screenshots below).

**Additional Notes**
- this PR introduces the concept of automatically instrumenting external libraries without requiring users to change configuration: Selenium should be required and configured by the time when test session starts, so I added [Contrib. auto_instrument_on_session_start!](https://github.com/DataDog/datadog-ci-rb/pull/183/files#diff-73e81d81ed4e915ac9a747cc6e0d9b23449ee88a47b6f0ad28c0f6cd9bd35c5dR13-R27) method that automatically instruments it when test session starts
- it turned out not to be enough to instrument `selenium-webdriver` gem because [capybara](https://github.com/teamcapybara/capybara) has its own logic to reset browser session between test runs. In this PR both `Selenium::WebDriver::Driver` and `Capybara::Selenium::Driver` (not that capybara is de-facto the default testing library for e2e tests in Ruby)

**How to test the change?**
Unit tests are provided.

Tested on DD staging using the following test:

```ruby
require_relative "helper"

class SeleniumTest < Minitest::Test
  include Capybara::DSL
  include Capybara::Minitest::Assertions

  def teardown
    Capybara.reset_sessions!
    Capybara.use_default_driver
  end

  def test_selenium
    visit "https://dd.datad0g.com/"
    assert page.has_content?("Datadog")
  end
end
```

With the following results

Browser tags and is_rum_active tag:
image

Browser session replay connected to a test:
image


lib/datadog/ci/contrib/selenium/ext.rb
```diff
@@ -0,0 +1,30 @@
...
+
+          SCRIPT_IS_RUM_ACTIVE = <<~JS
+            return !!window.DD_RUM
+          JS
+          SCRIPT_STOP_RUM_SESSION = <<~JS
+            if (window.DD_RUM && window.DD_RUM.stopSession) {
+              window.DD_RUM.stopSession();
+              return true;
+            } else {
+            	return false;
```
romainkomorndatadog, line 23: nit: indent?

lib/datadog/ci/contrib/selenium/rum.rb
```diff
@@ -0,0 +1,44 @@
...
+      module Selenium
+        # Provides functionality to interact with Datadog Real User Monitoring product
+        # via executing JavaScript code in the browser.
+        #
+        # Relevant docs: https://docs.datadoghq.com/real_user_monitoring/browser/
+        module RUM
+          def self.is_rum_active?(script_executor)
+            is_rum_active_script_result = script_executor.execute_script(Ext::SCRIPT_IS_RUM_ACTIVE)
+            Datadog.logger.debug { "[Selenium] SCRIPT_IS_RUM_ACTIVE result is #{is_rum_active_script_result.inspect}" }
+            is_rum_active_script_result == "true" || is_rum_active_script_result == true
```
romainkomorndatadog, line 20: Should the `script_executor` be ensuring its result's a bool, rather than checking for both here?

romainkomorndatadog (approved):

lib/datadog/ci/contrib/selenium/rum.rb
anmarchenko: script_executor is coming from selenium-webdriver, it could be `Selenium::WebDriver::Manager` or `Selenium::WebDriver::Driver`

One could say that we need to be exactly sure what type this returns but in Ruby case one can never be sure. Another more beatiful way to write it would be `is_rum_active_script_result.to_s == "true"`, I think this would look better

romainkomorndatadog: hehe, I would be more likely to try to cast the string to a bool :) but anyway, it's not very important, I think the code's fine as is.

PR #185, opened 2024-05-29
[NO-TICKET] automatically trace with correct time even when time is stubbed by timecop

anmarchenko:
**What does this PR do?**
timecop is a popular gem that is used to mock time in tests. It works by [replacing Time.now method](https://github.com/travisjeffery/timecop/blob/v0.9.8/lib/timecop/time_extensions.rb) and aliasing the original as `now_without_mock_time`.

This creates issues with tests tracing as reported spans has invalid time and thus are being skipped or reported with wrong timestamps.

It is possible to setup time_now_provider in settings: we can do this automatically when activating CI mode so that no manual work from user's side is needed.

**Motivation**
We received multiple customer support issues related to invalid start_time.

**How to test the change?**
Unit tests are provided (with timecop dependency).


anmarchenko:
Fun fact: it broke my tests that use Timecop as well! this means it works

juan-fernandez (approved):

PR #188, opened 2024-06-03
[NO TICKET] test code coverage with symlinked dependency required via entrypoint

anmarchenko:
**What does this PR do?**
This PR adds a test to assert that code coverage tool gets correct file paths when dependencies are located in symlinked folder and required via single entrypoint (emulating usage of local dependencies that are just symlinks to modules in minirepo).

**Motivation**
Make sure that code coverage won't break in the future as this is known case for the monorepo setup.

**How to test the change?**
This PR adds only unit tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -103,7 +103,7 @@ namespace :test do
           "bundle exec appraisal #{ruby_runtime}-#{appraisal_group} rake #{spec_task}"
         end
 
-        command += "'[#{spec_arguments}]'" if spec_arguments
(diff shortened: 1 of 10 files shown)


Rakefile
```diff
@@ -103,7 +103,7 @@ namespace :test do
           "bundle exec appraisal #{ruby_runtime}-#{appraisal_group} rake #{spec_task}"
```
anmarchenko, line 103: changes in this file are unrelated, just enjoying our own Datadog's static code analysis tool

nikita-tkachenko-datadog (approved):

PR #189, opened 2024-06-05
[SDTEST-408] multi threaded code coverage support for datadog_cov

anmarchenko:
**What does this PR do?**
Fixes an issue when per test code coverage for ITR ignored background threads. This led to the coverage being incomplete in the most cases as Rails uses threading for background jobs testing for example. 

Now by default code coverage is collected for all threads, not only the current thread of the test. I added a parameter that will allow the library to switch back to track single thread in a less probable situation where threaded parallel test runner is used.

**Motivation**
Reports of incomplete code coverage when running Rails tests

**Additional notes**
This is built on top of PR that improves code coverage tool performance: https://github.com/DataDog/datadog-ci-rb/pull/171

**How to test the change?**
Unit tests are provided


ext/datadog_cov/datadog_cov.c
```diff
@@ -88,9 +95,25 @@ static VALUE dd_cov_initialize(int argc, VALUE *argv, VALUE self)
   }
   VALUE rb_ignored_path = rb_hash_lookup(opt, ID2SYM(rb_intern("ignored_path")));
 
+  VALUE rb_threading_mode = rb_hash_lookup(opt, ID2SYM(rb_intern("threading_mode")));
+  int threading_mode;
+  if (!RTEST(rb_threading_mode) || rb_threading_mode == ID2SYM(rb_intern("multi")))
+  {
+    threading_mode = MULTI_THREADED_COVERAGE_MODE;
+  }
```
ivoanjo, line 103: Really minor: Consider maybe removing the `RTEST` and allowing it to fall down into the exception. My thinking is, it's kinda weird for `threading_mode: false` to result in the threading mode being activated... :thinking: 

```suggestion
  if (rb_threading_mode == ID2SYM(rb_intern("multi")))
  {
    threading_mode = MULTI_THREADED_COVERAGE_MODE;
  }
```

ext/datadog_cov/datadog_cov.c
```diff
@@ -154,31 +177,43 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
...
-  rb_thread_remove_event_hook(thval, dd_cov_update_coverage);
-
   struct dd_cov_data *dd_cov_data;
   TypedData_Get_Struct(self, struct dd_cov_data, &dd_cov_data_type, dd_cov_data);
 
+  if (dd_cov_data->threading_mode == SINGLE_THREADED_COVERAGE_MODE)
+  {
+    VALUE thval = rb_thread_current();
+    rb_thread_remove_event_hook(thval, dd_cov_update_coverage);
+  }
```
ivoanjo, line 214: So I guess this bug already existed, but it occurred to me that `stop` is incorrect in single-threaded mode if it accidentally gets called on another thread. Aka: maybe store the thread you're currently targeting? (And maybe `raise` if anyone tries to start more than once in single threaded mode?)

sig/datadog/ci/itr/coverage/ddcov.rbs
```diff
@@ -3,8 +3,9 @@ module Datadog
     module ITR
       module Coverage
         class DDCov
+          type threading_mode = :multi | :single
 
-          def initialize: (root: String, ignored_path: String?) -> void
+          def initialize: (root: String, ignored_path: String?, ?threading_mode: threading_mode?) -> void
```
ivoanjo, line 8: Minor: Maybe consider not making this argument optional? (The thinking being: if the production code always sets it, why make it more flexible?)

ivoanjo (commented):
:+1: Seems reasonable!

Given that dd-trace-rb also has a bunch of background threads, would it make sense to exclude them by default?

E.g. by looking at their names for instance (all of our threads should be named with the `Datadog::` prefix).

ext/datadog_cov/datadog_cov.c
anmarchenko: a-ha, you are right, it is kind of alleviated in the client code as DDCov is created and stored by thread, but I should make it more obvious

PR #197, opened 2024-07-05
[SDTEST-523] Expand test impact analysis with allocation tracing

anmarchenko:
**Problem statement**
Using line coverage for test impact analysis has a major limitation in Ruby: consider the following examples:

```ruby
# this class does not have any executable lines
class MyClass < OtherClass
end

test "instantiate MyClass" do
  assert MyClass.new != nil
end
```

The test "instantiate MyClass" does not cover MyClass because there are no executable lines in MyClass. If initializer was inherited from OtherClass, then this test will have `other_class.rb` in the list of covered files but not `my_class.rb`. 

This leads to a major intelligent test runner bug: if we change initializer of MyClass like that:
```ruby
class MyClass
  def initialize(arg)
    @arg = arg
  end
end
```

then the test above will start failing because `MyClass.new` expects argument now. But because `my_class.rb` is not covered by this test, intelligent test runner will skip test by default! It causes broken tests to be merged in the default branch.

If this example might seem artificial, unfortunately the same happens with ActiveRecord models or with ActiveModel classes:

```ruby
# perfectly valid Rails model that is not covered by any test
class Account < ApplicationRecord
  belongs_to :user
end
```

**Solution**
We cannot overcome this limitation by using line coverage: the code coverage approach works correctly in this case and this is just how line coverage works. We need to go deeper in Ruby VM tracing using techniques that are already used by continuous profiler. 

For this limitation, I've chosen to reach out for heap allocation tracepoint: it is possible to spy on every new object allocation that happens in Ruby heap. Even if no code from this class is executed during the test, it is enough for us to know that the test instantiates instances of this class to add its filename to the list of impacted files.

Notes on implementation:
- `RUBY_INTERNAL_EVENT_NEWOBJ` event type is used
- `rb_tracepoint_new` to register `RUBY_INTERNAL_EVENT_NEWOBJ` tracepoint
- `Module.const_source_location(klass_name)` Ruby API is used to get the source of a constant (every class name is a constant). This API is available from Ruby 2.7 - this is exactly the oldest Ruby version supported for test visibility product. We use this API from C with `rb_funcall(rb_cObject, rb_intern("const_source_location"), 1, klass_name)`
- `rb_rescue2` to safely get source code location of a class (it fails for many internal classes)
- Many libraries generate anonymous classes with class name like `#<Class:0ff0eabcde>` - we explicitly ignore them because getting source location for these classes always fails

**Known limitations**
- This approach only tracks objects instantiated during the test itself (in `before` hook or during the test). If the test suite has some models cached in global state and shared between tests, and these models don't have any methods implemented on them, the test impact analysis will still miss them. We need to change docs on Intelligent test runner in Ruby reiterating that **global state is harmful** and can cause flakiness and incompatibility with ITR.
- Ruby versions 3.2.0 - 3.2.2 have a bug that causes failures when using this tracepoint: allocation profiling is disabled for these versions
- `rb_tracepoint_new` cannot be attached to a specific thread, so the allocation tracing is enabled for multi threaded coverage mode only (this mode is the default one and the only one that can work for Rails, so it is not a major problem)
- Anonymous classes source locations are not supported

**How to test the change?**
Tested using by running test suites of the following open source projects:
- feedbin
- rubocop
- jekyll
- middleman
- vagrant
- devdocs

Unit tests reproducing the original problem are provided. See performance evaluation below.

**Performance evaluation**
Median performance overhead for some OSS projects' test suites before this change:
image

Results from benchmarks after this change:
- rubocop - [111,7% overhead](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/571084742) - 28% increase
- jekyll - [0,4% overhead](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/568646312) - 0% increase
- middleman - [37,6% overhead](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/571083432) - 32% increase
- devdocs - [23,8% overhead](https://gitlab.ddbuild.io/DataDog/apm-reliability/test-environment/-/jobs/571095569) - 39% increase

Overall this change increases code coverage overhead on test suites by 30-40% (compared to overhead before the change) in relative numbers. In absolute numbers depending on project's size and characteristics it means from 7% up to 30% more time spent in tests (the maximum overhead is from rubocop which is particularly challenging for profiling: 21k relatively fast tests).


lib/datadog/ci/configuration/components.rb
```diff
@@ -118,6 +110,42 @@ def activate_ci!(settings)
...
+            )
+
+            settings.ci.itr_test_impact_analysis_use_allocation_tracing = false
+          end
+
+          if RUBY_VERSION.start_with?("3.2.") && RUBY_VERSION < "3.2.3" &&
+              settings.ci.itr_test_impact_analysis_use_allocation_tracing
+            Datadog.logger.warn(
+              "Intelligent test runner: Allocation tracing is not supported in Ruby versions 3.2.0, 3.2.1 and 3.2.2 and will be forcibly " \
+              "disabled. This is due to a VM bug that can lead to crashes (https://bugs.ruby-lang.org/issues/19482). " \
```
devinburnette, line 130: 😭

anmarchenko: are you on 3.2.0 - 3.2.2?

lib/datadog/ci/configuration/components.rb
```diff
@@ -118,6 +110,42 @@ def activate_ci!(settings)
...
 
+        def build_test_optimisation(settings, test_visibility_api)
+          if settings.ci.itr_code_coverage_use_single_threaded_mode &&
+              settings.ci.itr_test_impact_analysis_use_allocation_tracing
+            Datadog.logger.warn(
+              "Intelligent test runner: Single threaded coverage mode is incompatible with allocation tracing. " \
+              "Allocation tracing will be disabled. It means that test impact analysis will not be able to detect " \
+              "instantiations of objects in your code, which is important for ActiveRecord models. " \
+              "Please add your app/model folder to the list of tracked files or disable single threaded coverage mode."
+            )
```
ivoanjo, line 121: If you ever want to support this -- you can use `rb_thread_current()` inside the tracepoint to compare it to 'th_covered` and ignore any events from threads that aren't your target.

ext/datadog_cov/datadog_cov.c
```diff
@@ -18,30 +27,75 @@ char *ruby_strndup(const char *str, size_t size)
...
+  // Line tracepoint can work in two modes: single threaded and multi threaded
+  //
+  // In single threaded mode line tracepoint will only cover the thread that started the coverage.
+  // This mode is useful for testing frameworks that run tests in multiple threads.
+  // Do not use single threaded mode for Rails applications unless you know that you
+  // don't run any background threads.
+  //
+  // In multi threaded mode line tracepoint will cover all threads. This mode is enabled by default
+  // and is recommended for most applications.
+  int threading_mode;
```
ivoanjo, line 79: Minor: I don't use them very often myself (yet...?) but this may be a good candidate for using an `enum`.

ext/datadog_cov/datadog_cov.c
```diff
@@ -71,64 +127,48 @@ const rb_data_type_t dd_cov_data_type = {
...
   dd_cov_data->root = NULL;
   dd_cov_data->root_len = 0;
   dd_cov_data->ignored_path = NULL;
   dd_cov_data->ignored_path_len = 0;
   dd_cov_data->last_filename_ptr = 0;
   dd_cov_data->threading_mode = MULTI_THREADED_COVERAGE_MODE;
 
-  return obj;
+  dd_cov_data->allocation_tracing_enabled = true;
+  dd_cov_data->object_allocation_tracepoint = rb_tracepoint_new(Qnil, RUBY_INTERNAL_EVENT_NEWOBJ, process_newobj_event, (void *)dd_cov);
```
ivoanjo, line 141: Minor: While I don't quite do that for the profiler, you could probably collapse these two into one, and just have `object_allocation_tracepoint == Qnil` when allocation is disabled, and containing the tracepoint when enabled.

ext/datadog_cov/datadog_cov.c
```diff
@@ -161,26 +201,140 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
...
+
+  // careful: this part of tracepoint is running for each allocated object, make it fast
+  rb_trace_arg_t *tracearg = rb_tracearg_from_tracepoint(tracepoint_data);
+  VALUE new_object = rb_tracearg_object(tracearg);
+
+  enum ruby_value_type type = rb_type(new_object);
+  if (type != RUBY_T_OBJECT && type != RUBY_T_STRUCT && type != RUBY_T_DATA)
   {
     return;
   }
```
ivoanjo, line 235: Note that excluding things such as `RUBY_T_HASH` or `RUBY_T_FILE` will also exclude their subclasses. Which may be an explicit decision...

In any case, it may be worth documenting here why others aren't being considered ATM.

ext/datadog_cov/datadog_cov.c
```diff
@@ -161,26 +201,140 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
...
+    return;
+  }
+
+  VALUE filename = RARRAY_AREF(source_location, 0);
+  if (filename == Qnil)
+  {
+    return;
+  }
+
+  record_impacted_file(dd_cov_data, filename);
```
ivoanjo, line 280: As per our discussion of "how much to do inside the tracepoint", have you considered perhaps recording only the observed class (specifically, `klass`), and only separately turning it into a file name? Assuming that folks aren't going to create a gazillion classes per single test case, holding on to all the class objects is probably not a lot of overhead.

ivoanjo (commented):
Left a few notes! I didn't do a pass very deep into the `get_source_location` part (ran out of time -- I can do it in the next pass :D) but hopefully this helps

ext/datadog_cov/datadog_cov.c
anmarchenko: Mostly for performance reasons I was going to filter for T_OBJECT only, but then I thought that stricter and Data objects have their fair share of usage in typical Rails apps - not sure how often people extend Hash or File in their projects. I will add it in the comment.

ext/datadog_cov/datadog_cov.c
anmarchenko: I think this is a great idea. I would gladly use rb_hash for it: not sure if this part of Ruby API is forbidden to use in trace point too?..

anmarchenko: I got reports of segfaults because of this approach:

```
/usr/local/ruby/lib/ruby/3.2.0/openssl/ssl.rb:278: [BUG] unexpected situation - recordd:1 current:0
ruby 3.2.3 (2024-01-18 revision 52bb2ac0a6) [x86_64-linux]
```

100% that I need to make the tracepoint code as small as possible, will work on that

lib/datadog/ci/configuration/components.rb
anmarchenko: I will hold off this for now as I am not sure if it's allowed inside the tracepoint ;)

devinburnette:
just tested this branch as of the latest commit (0b1618c) and verified the segfault is gone and the test impact analysis is working as expected for models.

anmarchenko:
Thank you for testing and your feedback @devinburnette! I'll do a couple more passes on Monday and will release soon

anmarchenko:
@ivoanjo this is ready for another pass - seems to be working now

lib/datadog/ci/configuration/components.rb
ivoanjo: Indeed, up to you! (I do suspect it's safe; as an alternative, you could always do some magic with thread-locals, e.g. by identifying which native thread matches the Ruby thread and then safely check in which native thread you landed during the hook)

ext/datadog_cov/datadog_cov.c
```diff
@@ -161,26 +210,152 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
...
+
+  if (RTEST(rb_ignored_path))
+  {
+    dd_cov_data->ignored_path_len = RSTRING_LEN(rb_ignored_path);
+    dd_cov_data->ignored_path = ruby_strndup(RSTRING_PTR(rb_ignored_path), dd_cov_data->ignored_path_len);
+  }
+
+  if (rb_allocation_tracing_enabled == Qtrue)
+  {
+    dd_cov_data->object_allocation_tracepoint = rb_tracepoint_new(Qnil, RUBY_INTERNAL_EVENT_NEWOBJ, on_newobj_event, (void *)self);
```
ivoanjo, line 350: Minor optimization: I believe you can pass in the `dd_cov_data` here directly, which would mean you could skip the `TypedData_Get_Struct` inside the hook.

ext/datadog_cov/datadog_cov.c
```diff
@@ -161,26 +210,152 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
...
-  // if the current filename is not located under the root, we skip it
-  if (strncmp(dd_cov_data->root, filename_ptr, dd_cov_data->root_len) != 0)
+  record_impacted_file(dd_cov_data, filename);
+}
+
+// Get source location for a given class name
+static VALUE get_source_location(VALUE klass_name)
+{
+  return rb_funcall(rb_cObject, rb_intern("const_source_location"), 1, klass_name);
+}
```
ivoanjo, line 220: So on the topic of performance, it occurs to me that if you're starting and stopping coverage on every individual test case, you may be redoing these kinds of lookups again and again.

Maybe not for this PR, but perhaps it's worth considering having some kind of a cache that would live across coverage start/stops? (Of course with proper sizing, etc)

ext/datadog_cov/datadog_cov.c
```diff
@@ -71,64 +137,47 @@ const rb_data_type_t dd_cov_data_type = {
...
   dd_cov_data->root_len = 0;
   dd_cov_data->ignored_path = NULL;
   dd_cov_data->ignored_path_len = 0;
   dd_cov_data->last_filename_ptr = 0;
-  dd_cov_data->threading_mode = MULTI_THREADED_COVERAGE_MODE;
+  dd_cov_data->threading_mode = multi;
 
-  return obj;
-}
+  dd_cov_data->object_allocation_tracepoint = Qnil;
```
ivoanjo, line 150: Minor: As we're discussing on chat, I'm not convinced that we need to avoid initializing things here, but this change is small/harmless so I think it's reasonable to go with this change and leave that debate separate.

ivoanjo (approved):
Left another round of suggestions, but I think overall my comments fall in the "extra things" bucket, and this PR seems reasonable even as-is.

ext/datadog_cov/datadog_cov.c
anmarchenko: Yes, interesting thing that I tried two kind of caches:
- hashtable (st.c) that tracks source for every class VALUE
- hashtable that for every class VALUE stores boolean value indicating whether the source file is from inside project or from external gem

Every combination of these 2 approaches ether do not change overhead for rubocop test source or makes it on about 4% slower (on average)! I did not investigate it further so I don't know yet why intuition does not work in this case: maybe if I have time I'll do another pass. 

For now, I am leaving it as is without additional caches as I believe that every optimisation must be backed by solid data, otherwise it will be just another source of bugs

ext/datadog_cov/datadog_cov.c
```diff
@@ -161,26 +210,152 @@ static void dd_cov_update_coverage(rb_event_flag_t event, VALUE data, VALUE self
...
-  // this is useful for ignoring bundled gems location
-  if (dd_cov_data->ignored_path_len != 0 && strncmp(dd_cov_data->ignored_path, filename_ptr, dd_cov_data->ignored_path_len) == 0)
+  VALUE klass = rb_class_of(new_object);
+  if (klass == Qnil || klass == 0)
+  {
+    return;
+  }
+  // Skip anonymous classes starting with "#<Class".
+  // it allows us to skip the source location lookup that will always fail
+  const char *name = rb_obj_classname(new_object);
```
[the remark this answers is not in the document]
anmarchenko: I like your `rb_mod_name` - it returns `Qnil` if klass is anonymous, so it works for filtering out anonymous classes, and works a bit faster

ext/datadog_cov/datadog_cov.c
ivoanjo: Sounds reasonable :+1:

PR #202, opened 2024-07-26
[SDTEST-116] Implement manual_api_events metric

anmarchenko:
**What does this PR do?**
Tracks manual_api_events metric when public API is used.

Internal instrumentations don't use public API anymore.

**How to test the change?**
Unit tests are provided

### DIFF
--- a/lib/datadog/ci/contrib/cucumber/formatter.rb
+++ b/lib/datadog/ci/contrib/cucumber/formatter.rb
@@ -35,14 +35,14 @@ module Datadog
           end
 
           def on_test_run_started(event)
-            CI.start_test_session(
+            test_visibility_component.start_test_session(
               tags: {
                 CI::Ext::Test::TAG_FRAMEWORK => Ext::FRAMEWORK,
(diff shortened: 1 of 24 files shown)


lib/datadog/ci/contrib/cucumber/formatter.rb
```diff
@@ -70,7 +70,7 @@ def on_test_case_started(event)
 
             start_test_suite(test_suite_name) unless same_test_suite_as_current?(test_suite_name)
 
-            test_span = CI.start_test(
```
nikita-tkachenko-datadog, line 73: I'm not sure what the difference is between `start_test` and `trace_test` (other than the `&block` parameter, which does not seem to be used here), is this change intentional?

nikita-tkachenko-datadog (approved):

lib/datadog/ci/contrib/cucumber/formatter.rb
anmarchenko: Yes, the `start_test` is a convenience method that does not take a code block. I implemented it for public API, but not for private API. Now because of telemetry I cannot use public API anymore, so no convenience for me :(

PR #205, opened 2024-07-26
[SDTEST-160] git commands telemetry

anmarchenko:
**What does this PR do?**
Adds telemetry metrics collection to git commands

**Additional Notes**
This change caused curious fail: one must not access Telemetry when the library is initialized. We previously collected environment tags during the library initialization stage, which might include using git commands. I made the environment tags collection lazy, so it does not happen until at least one span is traced.

**How to test the change?**
Unit tests are provided

### DIFF
--- a/lib/datadog/ci/git/local_repository.rb
+++ b/lib/datadog/ci/git/local_repository.rb
@@ -3,12 +3,25 @@
 require "open3"
 require "pathname"
 
+require_relative "../ext/telemetry"
+require_relative "telemetry"
 require_relative "user"
 
 module Datadog
   module CI
     module Git
       module LocalRepository
+        class GitCommandExecutionError < StandardError
+          attr_reader :output, :command, :status
+          def initialize(message, output:, command:, status:)
+            super(message)
+
+            @output = output
+            @command = command
+            @status = status
+          end
+        end
+
         COMMAND_RETRY_COUNT = 3
 
         def self.root
@@ -51,6 +64,7 @@ module Datadog
           exec_git_command("git ls-remote --get-url")
         rescue => e
           log_failure(e, "git repository url")
+          telemetry_track_error(e, Ext::Telemetry::Command::GET_REPOSITORY)
           nil
         end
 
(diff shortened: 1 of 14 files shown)


lib/datadog/ci/git/local_repository.rb
```diff
@@ -48,9 +61,18 @@ def self.current_folder_name
         end
 
         def self.git_repository_url
-          exec_git_command("git ls-remote --get-url")
+          Telemetry.git_command(Ext::Telemetry::Command::GET_REPOSITORY)
+          res = nil
+
+          duration_ms = Core::Utils::Time.measure(:float_millisecond) do
+            res = exec_git_command("git ls-remote --get-url")
```
romainkomorndatadog, line 68: Small question: why not have `exec_git_command()` do the measuring and return some kind of tuple with result, duration, and status code? Or alternatively, you could pass the git command as a parameter and have the underlying method do the telemetry piece.

It would be a very modest win in code deduplication.

romainkomorndatadog (approved):

lib/datadog/ci/git/local_repository.rb
anmarchenko: I'm not a fan of DRY in general and I eager to accept code duplication every time over confusing behaviour.

Changing return of the method to include telemetry mixes 2 different concepts together: executing command and measuring it. If I return back to this code after 2 months, I will never remember why `exec_git_command` returns a tuple. What's worse, we measure only subset of git commands (for whatever reason). This means that these return values will be used in some cases but ignored in other and it will be absolutely impossible to explain why. 

In this case deduplicating the code would completely break `exec_git_command` abstraction: I will take duplication over broken abstraction every day (like Sandi taught me! :) )

romainkomorndatadog: This is why I added [a wrapper around `_git_subprocess()`](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/ext/git.py#L83-L109)  in `dd-trace-py` and only call it when I'm going to use the duration and status code for telemetry.  😄

romainkomorndatadog: I ended up [inserting a wrapped function](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/ext/git.py#L83-L109) for the Python tracer... but since I had to maintain backwards compatibility, I kept the original function as a wrapper of the new one, which may or may not have been the right move.

anmarchenko: I'm fine with duplication for now, but I will refactor if it'll become painful in the future

PR #207, opened 2024-07-30
[SDTEST-415] Add test_session metric

anmarchenko:
**What does this PR do?**
Adds test_session metric with provider tag to track which CI providers are used

**How to test the change?**
Unit tests

### DIFF
--- a/lib/datadog/ci/test_visibility/telemetry.rb
+++ b/lib/datadog/ci/test_visibility/telemetry.rb
@@ -18,6 +18,23 @@ module Datadog
           Ext::AppTypes::TYPE_TEST_SESSION => Ext::Telemetry::EventType::SESSION
         }.freeze
 
+        PROVIDER_TAG_TO_TELEMETRY_PROVIDER_TAG = {
+          Ext::Environment::Provider::APPVEYOR => Ext::Telemetry::Provider::APPVEYOR,
+          Ext::Environment::Provider::AWS => Ext::Telemetry::Provider::AWS,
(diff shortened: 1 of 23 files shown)


lib/datadog/ci/test_visibility/telemetry.rb
```diff
@@ -18,6 +18,23 @@ module Telemetry
           Ext::AppTypes::TYPE_TEST_SESSION => Ext::Telemetry::EventType::SESSION
         }.freeze
 
+        PROVIDER_TAG_TO_TELEMETRY_PROVIDER_TAG = {
```
anmarchenko, line 21: This is incredibly ugly: for some reason we use different values for some providers when submitting to citestcycle and telemetry. For example Microsoft Azure is "azurepipelines" in citestcylce but "azp" in telemetry

romainkomorndatadog: Hot take: I think we shouldn't do this and I think we should fix the other tracers to use the same tags for telemetry and citestcycle.

Unless I'm mistaken, the backend doesn't have an enum validating these values, so worse case, we'd have some extra values for some time.

romainkomorndatadog (approved):

lib/datadog/ci/test_visibility/telemetry.rb
anmarchenko: I like your proposal, but I also would like to move fast and leave this epic behind. Given that the discussion about naming can go on for months from now, let's merge this one as is and then I'll fix it in one of the subsequent PRs

PR #213, opened 2024-08-07
[SDTEST-437] Auto test retries for RSpec

anmarchenko:
**What does this PR do?**
Adds [auto test retries](https://docs.datadoghq.com/tests/auto_test_retries) functionality for RSpec test framework

**How to test the change?**
Tested using middleman test suite with this flaky test: 
```ruby
    it "is new flaky" do
      res = rand(3)
      expect(res).to eq 1
    end
```

Locally test process completed normally:
image

**Results of the test run in Datadog**

New flaky test detected:
image

All the retries are correctly displayed in Datadog test runs:
image

Test session has status "pass":
image


lib/datadog/ci/contrib/rspec/example.rb
```diff
@@ -37,47 +37,66 @@ def run(*args)
...
+              # if test passed at least once
+              finish(reporter)
+            end
+
+            def finish(reporter)
+              # by default finish test but do not report it to RSpec::Core::Reporter
+              # it is going to be reported once after retries are done
+              return super unless @skip_reporting
+
+              super(::RSpec::Core::NullReporter)
```
devinburnette, line 99: if we're delaying the reporting until after the retries, then isn't it true that we would potentially hide the fact that there were flakes? or will the final trace data include the flake information?

anmarchenko: This is hiding reporting (for now) to RSpec formatter that prints the final message and tracks tests that failed (I might fix it in later versions to format retries correctly).

As tested in instrumentation_spec every retry is traced as a separate test span, so the test will be correctly identified as flaky. Also, test suite and test session statuses will be correctly reported as pass.

juan-fernandez (approved):

PR #228, opened 2024-09-03
[SDTEST-756] remove monkey patch of core identity module

anmarchenko:
**What does this PR do?**
Removes monkey patching of Core::Identity module that was added to report datadog-ci version instead of datadog version

**Motivation**
To be replaced with https://github.com/DataDog/dd-trace-rb/pull/3881


ivoanjo (approved):
:+1: LGTM. Thanks for looking into this!

PR #238, opened 2024-09-23
[SDTEST-528] Add pull_request extra tags for GitHub Actions

anmarchenko:
**What does this PR do?**
Adds extra tags with Git metadata for Github Actions workflows that are triggered with pull_request event.

**Motivation**
In GHA pull_request workflows HEAD does not point to the last commit in the PR, but to a commit that is the result of merging PR branch with the target branch.

Datadog backend needs additional information in this case to correctly determine the latest commit in the PR branch.

**How to test the change?**
Unit tests are provided


spec/datadog/ci/contrib/rspec/instrumentation_spec.rb
```diff
@@ -967,7 +967,7 @@ def rspec_skipped_session_run
       let(:integration_options) { {service_name: "lspec"} }
     end
 
-    it "retries test until it passes" do
+    it "does not report the test that failed when RSpec was quitting" do
```
anmarchenko, line 970: This change is not relevant to this PR's topic, I just noticed that one of the specs was copy-pasted without changing description

spec/support/fixtures/ci/github.json
```diff
@@ -710,5 +710,70 @@
       "git.commit.sha": "b9f0fb3fdbb94c9d24b2c75b49663122a529e123",
       "git.repository_url": "https://1.1.1.1:1234/ghactions-repo.git"
     }
+  ],
+  [
```
anmarchenko, line 714: here are the actual test cases:
1. happy path
2. case where events file is not found

juan-fernandez (approved):

PR #244, opened 2024-10-01
[SDTEST-878] Optimise LocalRepository.relative_to_root helper to make test impact analysis faster

anmarchenko:
**What does this PR do?**
Applies some heuristics to avoid calling `Pathname#relative_path_from` function when determining what would be path relative to the git root for the given filepath:
- If given filepath is absolute, we use the fact that absolute paths are pre-filtered already and they always contain root as prefix, so we only need to slice the string to get the suffix (`path[prefix_length..]`)
- If given filepath is relative, we compute `Pathname#relative_path_from` once and then we store the difference between the relative filepath and the computed relative path from root in the `@prefix_to_root` cached variable. We can do it because all relative paths are computed relative to the `Dir.pwd` which doesn't change in the lifetime of the process. After we have `@prefix_to_root` cache, we prepend it to every filepath we get (or just return the filepath if the tests are running from the git root).

**Motivation**
Profiling rubocop test suite with Datadog continuous profiler showed that big part of CPU time is spent in `Datadog::CI::Git::LocalRepository.relative_to_root` function:
profile-before

**How to test the change?**
Unit tests are provided.

[Benchmark results](https://github.com/DataDog/test-environment/pull/329):
- Middleman overhead 34,1% (down from 40%)
- Rubocop overhead 64% (down from 128%) - **2x improvement**

### DIFF
--- a/lib/datadog/ci/transport/http.rb
+++ b/lib/datadog/ci/transport/http.rb
@@ -149,6 +149,10 @@ module Datadog
             nil
           end
 
+          def response_size
+            0
+          end
+
           def inspect
             "ErrorResponse error:#{error}"
           end
--- a/lib/datadog/ci/git/local_repository.rb
+++ b/lib/datadog/ci/git/local_repository.rb
@@ -36,10 +36,32 @@ module Datadog
           root_path = root
           return path if root_path.nil?
 
-          path = Pathname.new(File.expand_path(path))
-          root_path = Pathname.new(root_path)
+          if File.absolute_path?(path)
+            res = path.gsub(root, "")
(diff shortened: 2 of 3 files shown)


lib/datadog/ci/transport/http.rb
```diff
@@ -149,6 +149,10 @@ def code
             nil
           end
 
+          def response_size
+            0
```
anmarchenko, line 153: unrelated change, I noticed that library started to fail if certain backend API is not available because for ErrorResponse there is no http_response

juan-fernandez (approved):

PR #246, opened 2024-10-08
[SDTEST-744] add memory leaks tests with ruby_memcheck

anmarchenko:
**What does this PR do?**
Adds https://github.com/Shopify/ruby_memcheck gem to automatically find and report memory leaks using [valgrind](https://valgrind.org/). This PR mostly copies the setup from dd-trace-rb which was done in https://github.com/DataDog/dd-trace-rb/pull/3852

**Motivation**
Prevent memory leaks from the native extension

**How to test the change?**
Github Actions workflow is added

### DIFF
--- a/.github/workflows/test-memory-leaks.yml
+++ b/.github/workflows/test-memory-leaks.yml
@@ -1,0 +1,15 @@
+name: Test for memory leaks
+on: [push]
+jobs:
+  test-memcheck:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.4.0-preview1 # TODO: Use stable version once 3.4 is out
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+          bundler: latest
+          cache-version: v1 # bump this to invalidate cache
+      - run: sudo apt install -y valgrind && valgrind --version
+      - run: bundle exec rake compile spec:ddcov_memcheck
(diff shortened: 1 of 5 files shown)


.github/workflows/test-memory-leaks.yml
```diff
@@ -0,0 +1,15 @@
...
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.4.0-preview1 # TODO: Use stable version once 3.4 is out
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+          bundler: latest
+          cache-version: v1 # bump this to invalidate cache
+      - run: sudo apt install -y valgrind && valgrind --version
```
ivoanjo, line 14: You may want to borrow some of the tweaks we've done since to avoid flakiness in installing valgrind: https://github.com/DataDog/dd-trace-rb/blob/master/.github/workflows/test-memory-leaks.yaml#L14

suppressions/ruby-3.4.supp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+# We don't care about the tr external tool
+{
+  pkg-config-memory
+  Memcheck:Leak
+  ...
+  obj:/usr/bin/tr
+  ...
+}
```
ivoanjo, line 53: I suspect you won't need these, but also they're harmless to keep, if you want to just keep this file mostly in sync with dd-trace-rb

ivoanjo (approved):
:+1: LGTM

PR #252, opened 2024-11-05
[SDTEST-1129] Internal: decouple patcher from tracing, update ruby 3.4 compatibility

anmarchenko:
**What does this PR do?**
Migrates patcher code from Datadog::Tracing::Contrib to Datadog::CI::Contrib

**Motivation**
Reduce coupling with private tracing details. Move all contrib/integrations code to CI to be able to develop independent auto-instrumentation feature.

**How to test the change?**
It works if unit tests still pass


sig/datadog/ci/contrib/patcher.rbs
```diff
@@ -0,0 +1,33 @@
...
+
+          def name: () -> String
+
+          def patched?: () -> bool
+
+          def patch: () -> void
+
+          def on_patch_error: (untyped e) -> untyped
+
+          def default_tags: () -> untyped
```
Strech, line 22: I know that it might be cumbersome to fill in, but maybe we can shape it a bit? Like if it's Array, we can state it?

Strech (approved):
LGTM 👍🏼 

If you have time, maybe it makes sense to give a bit love to the untyped RBS returns.

PR #260, opened 2024-11-26
[SDTEST-228] Send auto_injected: true to internal telemetry when auto instrumentation was used

anmarchenko:
**What does this PR do?**
Adds a couple of small fixes to auto instrumentation:
- `test_session` telemetry metric has now auto_injected tag set to true when auto instrumentation was used
- when auto instrumentation is used on JRuby, we print error message and bail out of instrumenting tests

### DIFF
--- a/lib/datadog/ci/contrib/instrumentation.rb
+++ b/lib/datadog/ci/contrib/instrumentation.rb
@@ -9,6 +9,7 @@ module Datadog
         class InvalidIntegrationError < StandardError; end
 
         @registry = {}
+        @auto_instrumented = false
 
         def self.registry
           @registry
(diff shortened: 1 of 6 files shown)


lib/datadog/ci/contrib/instrumentation.rb
```diff
@@ -18,13 +19,18 @@ def self.register_integration(integration_class)
...
+        end
+
         # Auto instrumentation of all integrations.
         #
         # Registers a :script_compiled tracepoint to watch for new Ruby files being loaded.
         # On every file load it checks if any of the integrations are patchable now.
         # Only the integrations that are available in the environment are checked.
         def self.auto_instrument
           Datadog.logger.debug("Auto instrumenting all integrations...")
+          @auto_instrumented = true
```
romainkomorndatadog, line 33: Should this be after
```
if auto_instrumented_integrations.empty?
```
since you can end up running without instrumentation?
?

romainkomorndatadog (approved):

PR #275, opened 2025-01-02
[NOTICKET-1] bump maximum Ruby version to 3.5

anmarchenko:
**What does this PR do?**
Bumps maximum Ruby version to 3.5

Closes https://github.com/DataDog/datadog-ci-rb/issues/273

**Motivation**
Ruby 3.4 is released

**How to test the change?**
Ruby 3.4 is tested by unit tests since some time

PR #276, opened 2025-01-02
[NOTICKET-1] Bump to version 1.11.0

anmarchenko:
### Changed

* bump maximum Ruby version to 3.5 (#275)
* Use logical test session name as part of test session span's resource instead of test command (#271)

### Fixed

* set the max payload size for events to 4.5MB (#272)
* Fix inline comments handling when parsing CODEOWNERS files (#267)


ivoanjo (approved):
:+1: LGTM

PR #285, opened 2025-02-05
ci: pin all GitHub Actions by SHA and update via dependabot

xopham:
**What does this PR do?**

- **Add dependabot for github actions**
- **Pin actions by hash**

**Motivation**

Pinning 3rd-party GitHub Actions by commit SHA makes them less vulnerable to compromise of the 3rd party. To avoid outdating and non-verbosity, versions are commented after the SHA and updating via dependabot is introduced that will automatically update the commented version tag as well.

In case of a false commit SHA, this change could break the corresponding workflow. Typically, this does not cause major interruptions, but it can for example affect a release pipeline and require restart causing delays.

**Additional Notes**

**How to test the change?**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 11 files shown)


TonyCTHsu (approved):

anmarchenko (approved):

TonyCTHsu:
I made the changes in
- https://github.com/DataDog/datadog-ci-rb/pull/287
- https://github.com/DataDog/datadog-ci-rb/pull/288

PR #306, opened 2025-04-03
[SDTEST-426] agentless logs submission for tests

anmarchenko:
**What does this PR do?**
Allows logs to be submitted to Datadog without Datadog Agent when tracing test session. Logs generated by tests can then be explored in Datadog Logs product or on the Test Run page in Datadog Test Optimization.

This feature is off by default because it generates additional costs for the user, set DD_AGENTLESS_LOG_SUBMISSION_ENABLED to 1 to enable it.

This PR only adds support for `ActiveSupport::TaggedLogging` logger. 
In the subsequent PRs I plan to add support for `lograge` and `semantic_logger`. 

Note that this feature depend on APM log correlation already being setup correctly by customer (see [Ruby on Rails logs collection](https://docs.datadoghq.com/logs/log_collection/ruby/?tab=lograge) and [Correlating Ruby logs and traces](https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/ruby/))

**Motivation**
Planned feature work for Q2 2025

**How to test the change?**
Test using https://github.com/anmarchenko/quotes-rails test app, here is how logs are displayed:
image


lib/datadog/ci/async_writer.rb
```diff
@@ -0,0 +1,112 @@
...
+
+        @stopped = false
+      end
+
+      def write(event)
+        return if @stopped
+
+        # Start worker thread. If the process has forked, it will trigger #after_fork to
+        # reconfigure the worker accordingly.
+        perform
```
vitor-de-araujo, line 53: I'm curious to understand how this works, as I don't know much Ruby. What makes this call start a thread? `perform` below takes `*events` to send, but nothing is passed in this call.

vitor-de-araujo (approved):

lib/datadog/ci/async_writer.rb
anmarchenko: Oh, there are layers of layers of dynamic things in Datadog::Core for that! This API is absolutely not intuitive and I don't like it even a bit, but I guess it became historically like that.

Exact line of code that makes the thread start is here:
https://vscode.dev/github/DataDog/dd-trace-rb/blob/master/lib/datadog/core/workers/async.rb

But it took me a very long time to dig it out when I started to work with this library.

PR #327, opened 2025-05-21
[SDTEST-2040] calculate end lines for tests

anmarchenko:
**What does this PR do?**
Adds `test.source.end` tag to tests. This tag contains information about the end line for source code of each test.

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/ext/datadog_source_code/datadog_source_code.c
+++ b/ext/datadog_source_code/datadog_source_code.c
@@ -1,0 +1,26 @@
+#include <ruby.h>
+#include <ruby/debug.h>
+
+typedef struct rb_iseq_struct rb_iseq_t;
+const rb_iseq_t *rb_iseqw_to_iseq(VALUE iseqw);
+
+void rb_iseq_code_location(const rb_iseq_t *, int *first_lineno,
+                           int *first_column, int *last_lineno,
+                           int *last_column);
+
+static VALUE iseq_last_line(VALUE self, VALUE iseqw) {
+  const rb_iseq_t *iseq = rb_iseqw_to_iseq(iseqw);
+  int line;
+  rb_iseq_code_location(iseq, NULL, NULL, &line, NULL);
+  return INT2NUM(line);
+}
+
+void Init_datadog_source_code(void) {
+  VALUE mDatadog = rb_define_module("Datadog");
+  VALUE mCI = rb_define_module_under(mDatadog, "CI");
+  VALUE mUtils = rb_define_module_under(mCI, "Utils");
+  VALUE mSourceCode = rb_define_module_under(mUtils, "SourceCode");
+
+  rb_define_singleton_method(mSourceCode, "_native_last_line_from_iseq",
+                             iseq_last_line, 1);
+}
--- a/lib/datadog/ci/utils/source_code.rb
+++ b/lib/datadog/ci/utils/source_code.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Datadog
+  module CI
+    module Utils
+      module SourceCode
+        begin
+          require "datadog_source_code.#{RUBY_VERSION}_#{RUBY_PLATFORM}"
+
+          LAST_LINE_AVAILABLE = true
+        rescue LoadError
+          LAST_LINE_AVAILABLE = false
+        end
+
+        def self.last_line(proc)
+          return nil unless LAST_LINE_AVAILABLE
+
+          # Ruby has outdated RBS for RubyVM::InstructionSequence where method `of` is not defined
+          # steep:ignore:start
+          iseq = RubyVM::InstructionSequence.of(proc)
+          # steep:ignore:end
+          _native_last_line_from_iseq(iseq)
+        end
+      end
+    end
+  end
+end
--- a/Rakefile
+++ b/Rakefile
@@ -209,6 +209,10 @@ Rake::ExtensionTask.new("datadog_cov.#{RUBY_VERSION}_#{RUBY_PLATFORM}") do |ext|
   ext.ext_dir = "ext/datadog_cov"
 end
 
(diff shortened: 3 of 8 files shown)


ext/datadog_source_code/datadog_source_code.c
```diff
@@ -0,0 +1,26 @@
...
+void rb_iseq_code_location(const rb_iseq_t *, int *first_lineno,
+                           int *first_column, int *last_lineno,
+                           int *last_column);
+
+static VALUE iseq_last_line(VALUE self, VALUE iseqw) {
+  const rb_iseq_t *iseq = rb_iseqw_to_iseq(iseqw);
+  int line;
+  rb_iseq_code_location(iseq, NULL, NULL, &line, NULL);
+  return INT2NUM(line);
+}
```
ivoanjo, line 16: I was curious what would happen if the iseq was somehow not valid and indeed Ruby is assuming it's indeed an iseq and YOLO'ing here:

```
3.2.5 :002 > require_relative 'lib/datadog/ci/utils/source_code.rb'
 => true 
3.2.5 :003 > Datadog::CI::Utils::SourceCode.last_line(123)
x-datadog-ci-rb/lib/datadog/ci/utils/source_code.rb:23: [BUG] Segmentation fault at 0x0000000000000024
ruby 3.2.5 (2024-07-26 revision 31d0f1a2e7) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0024 p:---- s:0131 e:000130 CFUNC  :_native_last_line_from_iseq
```

So you'll probably want to do some kind of strict checking here that you're really holding on to an iseq.

lib/datadog/ci/contrib/minitest/test.rb
```diff
@@ -27,17 +28,25 @@ def before_setup
...
+              source_file, first_line_number = test_method.source_location
+              last_line_number = Utils::SourceCode.last_line(test_method)
+
+              # @type var tags : Hash[String, String]
+              tags = {
+                CI::Ext::Test::TAG_FRAMEWORK => Ext::FRAMEWORK,
+                CI::Ext::Test::TAG_FRAMEWORK_VERSION => datadog_integration.version.to_s,
+                CI::Ext::Test::TAG_SOURCE_FILE => Git::LocalRepository.relative_to_root(source_file),
+                CI::Ext::Test::TAG_SOURCE_START => first_line_number.to_s
+              }
```
ivoanjo, line 42: This is not really introduced in this PR, but it occured to me that `#source_location` can be nil sometimes, e.g.:

```
[].method(:each).source_location
 => nil 
```

...in practice it shouldn't happen on a test suite, but mentioning in case you may want to do a bit more defensive coding around this case?

lib/datadog/ci/test_optimisation/component.rb
ivoanjo, line 1: I'm somewhat curious about the introduction of synchronization on `@skippable_tests`. In particular, there's A LOT of other places in the file where `@skippable_tests` get accessed without the mutex.

It may be worth documenting what this mutex is protecting and when, as otherwise it can be a bit hard to get right in future modifications?

ivoanjo (commented):
Left a few notes, but this is :100:% a super cool trick! Really simple implementation, really impactful change. Gj! :)

ext/datadog_source_code/datadog_source_code.c
anmarchenko: great catch!

lib/datadog/ci/test_optimisation/component.rb
anmarchenko: yes, this one solves one rare edge case for JRuby (unrelated to other changes unfortunately), I might need to review and add more synchronization for this and also add a comment why is it needed

lib/datadog/ci/utils/source_code.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+
+        def self.last_line(target)
+          return nil if target.nil?
+          return nil unless LAST_LINE_AVAILABLE
+
+          # Ruby has outdated RBS for RubyVM::InstructionSequence where method `of` is not defined
+          # steep:ignore:start
+          iseq = RubyVM::InstructionSequence.of(target)
+          return nil unless iseq.is_a?(RubyVM::InstructionSequence)
```
ivoanjo, line 22: If possible, I think it would nice safer to implement this check using native code, so that there's no way to go around our checks and still cause damage?

Of course, there's other ways for a Ruby app to self-trigger a segfault if it wants to (`Fiddle.free(42)` or `Process.kill('SEGV', Process.pid)` come to mind), but, Idk, it seems a bit of a nice property that the method won't cause bad things, no matter what you can throw at it :eyes:

anmarchenko: that's turned out to be harder than I though

the only check I could find in MRI code for that is:
```
        if (CLASS_OF(op) == rb_cISeq) {
```

but `rb_cISeq` isn't exported anywhere

It is possible to get this class in C by using `rb_path_to_class` I guess, but then it might open another can of worms - is it really worth it?

ivoanjo: > but rb_cISeq isn't exported anywhere

Good point. I was looking it up and...

```ruby
void
Init_ISeq(void)
{
    /* declare ::RubyVM::InstructionSequence */
    rb_cISeq = rb_define_class_under(rb_cRubyVM, "InstructionSequence", rb_cObject);
```

it doesn't look like `rb_cIseq` is something very "weird".

This means I can think of two simple ways to get ahold of it:

1. Looking the class up in native code (e.g. `rb_const_get(rb_cObject, rb_intern("InstructionSequence"));`)
2. Asking for an iseq in native code and then grabbing the result of calling `CLASS_OF` on it.

PR #487, opened 2026-04-03
Use c.git.repository_url and c.git.commit_sha if it is defined instead of env vars

vpellan:
**What does this PR do?**

This PR replace the direct access to DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA to access to the config DSL. There is a fallback mechanism if the option is not defined on dd-trace-rb (for older versions).

**Motivation**

We're trying to move as much env vars as possible to the config DSL, so that they can be configured programmatically, but also through stable config. These two configs are common to both dd-trace-rb and datadog-ci-rb.

**Additional Notes**

Depends on https://github.com/DataDog/dd-trace-rb/pull/5559

**How to test the change?**


anmarchenko:
@vpellan closing this for now (almost 5 months passed) - feel free to reopen if it will be relevant again!

PR #540, opened 2026-07-15
Add DD_CIVISIBILITY_ITR_CODE_COVERAGE_ENABLED to allow test skipping without per-test coverage collection

daande:
## Summary

Adds `DD_CIVISIBILITY_ITR_CODE_COVERAGE_ENABLED` (setting `ci.itr_code_coverage_enabled`, default `true`) — a client-side opt-out for per-test code coverage collection in Test Impact Analysis. With the flag set to `false`, a session still applies backend-provided test skipping but never loads the native coverage extension or collects per-test coverage. Default `true` means no behavior change for existing users.

## Problem

Both `code_coverage_enabled` and `tests_skipping_enabled` come solely from backend library settings — there is no client-side way to apply skipping **without** collecting per-test coverage.

Coverage instrumentation makes each executed test measurably slower: on our large Rails suite (~63k tests) a matched-pairs experiment measured roughly **2x** per executed test. That creates a break-even problem: whenever the skippable percentage is below ~50%, an instrumented run that skips tests costs *more* total compute than an uninstrumented run of the full suite. Today the only way out is disabling ITR entirely, which also gives up the skipping.

## Solution

A single new AND-term in `TestImpactAnalysis::Component#configure`:

```ruby
@code_coverage_enabled = @enabled && remote_configuration.code_coverage_enabled? && @code_coverage_collection_enabled
```

- New env var `DD_CIVISIBILITY_ITR_CODE_COVERAGE_ENABLED` / setting `ci.itr_code_coverage_enabled` (bool, default `true`), wired through `Configuration::Components` into the component as `code_coverage_collection_enabled:`.
- Test skipping is untouched — skippable tests are still fetched and applied.
- When disabled, `load_datadog_cov!` is never called (the native coverage extension is not loaded), and the `test.code_coverage.enabled` session tag correctly reports `false`, so the backend can tell these sessions uploaded no coverage.
- RBS signatures updated for the new constant and constructor keyword.

This lets an org segment its CI: a dedicated lane (e.g. throttled default-branch runs) keeps the TIA coverage dataset fresh with full instrumentation, while PR sessions apply skips at ~1x per-test cost — making skipping a net win at any skippable percentage instead of only above the ~50% break-even.

**Question for maintainers:** is there any server-side expectation that skipping sessions also upload coverage (dataset freshness aside)? If so, happy to add a warning log or docs note. The flag defaults to `true`, so nothing changes unless explicitly opted in.

## Review Guide

**Call flow** — how the flag reaches the decision:

```
Configuration::Settings      (lib/datadog/ci/configuration/settings.rb)   ← option :itr_code_coverage_enabled, env-backed, default true
   └── Configuration::Components (lib/datadog/ci/configuration/components.rb) ← passes code_coverage_collection_enabled:
          └── TestImpactAnalysis::Component#configure (lib/datadog/ci/test_impact_analysis/component.rb) ← new AND-term on @code_coverage_enabled
```

**Focus areas:**
- **`lib/datadog/ci/test_impact_analysis/component.rb`** — the one behavioral line; everything downstream (`code_coverage?`, `load_datadog_cov!`, coverage start/stop, session tag) already branches on `@code_coverage_enabled`, so no other call sites change.

## Test plan

**Verifiable by agent before merging**
- [x] New settings specs: default `true`, env var `true`/`false`/unset, and setter (`spec/datadog/ci/configuration/settings_spec.rb`)
- [x] New component specs ("code coverage collection is disabled locally"): skipping stays enabled with skippable tests applied, `code_coverage?` is `false`, session tags reflect the override, and coverage never starts (`spec/datadog/ci/test_impact_analysis/component_spec.rb`)
- [x] Ran the three affected spec files (settings, component, components) after `rake compile_ext` — 245 examples, 0 failures (Ruby 3.4.7, arm64-darwin)
- [x] `bundle exec standardrb` clean on all changed files
- [x] `bundle exec rake steep:check` — output identical to a clean `main` checkout in my environment (pre-existing "library RBS file" load errors locally; zero errors attributable to this change)

**Cannot be verified before merge**
- [ ] End-to-end behavior against a live Datadog backend (a skipping session that uploads no coverage) — requires a TIA-enabled service with backend-provided skip sets


anmarchenko:
Hi @daande! 

I responded you in our public slack - when using ddtest and suite level skipping, you could use `DD_CIVISIBILITY_ITR_ENABLED=0` - there is no in-process skipping anyway in this case.

For now, I would not add this setting because it can be easily misused and make TIA unusable: our feature design assumes that test impact analysis is `always on`

anmarchenko:
closing this one for now

PR #570, opened 2026-08-11
Fix SimpleCov compatibility

### LINKED ISSUE #567: SimpleCov contrib breaks on simplecov main (post-1.0.3): undefined method 'add_not_loaded_files' for module SimpleCov
## Summary

Datadog::CI::Contrib::Simplecov::ResultExtractor::ClassMethods#__dd_peek_result calls add_not_loaded_files, a private method on the SimpleCov singleton. SimpleCov removed that method after v1.0.3, so the call raises NoMethodError from TestSession#finish, which aborts the RSpec runner and takes the whole test process down with it — including SimpleCov's own report.

## Versions

- datadog-ci 1.36.0
- simplecov main (post-v1.0.3; will ship in the next release)
- Ruby 4.0, CircleCI, RSpec

## Stack trace
```
bundler: failed to load command: rspec (/home/circleci/project/vendor/bundle/ruby/4.0.0/bin/rspec)
Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected
.../datadog-ci-1.36.0/lib/datadog/ci/contrib/simplecov/result_extractor.rb:27:in
  'Datadog::CI::Contrib::Simplecov::ResultExtractor::ClassMethods#__dd_peek_result':
  undefined method 'add_not_loaded_files' for module SimpleCov (NoMethodError)
  from .../datadog/ci/test_tracing/deprecated_total_coverage_metric.rb:25:in 'extract_lines_pct'
  from .../datadog/ci/test_tracing/component.rb:316:in 'on_test_session_finished'
  from .../datadog/ci/test_tracing/component.rb:191:in 'deactivate_test_session'
  from .../datadog/ci/test_session.rb:20:in 'finish'
  from .../datadog/ci/contrib/rspec/runner.rb:56:in 'run_specs'
```

## Root cause

SimpleCov moved unloaded-file injection out of each individual process and into the merge step (simplecov#1250). Previously each worker in a parallel run simulated nearly the whole project and the merge discarded all but one copy; now only the union of what every process loaded determines what was truly never loaded, so injection happens once in SimpleCov::ResultMerger.

As part of that, the private SimpleCov.add_not_loaded_files(result) — which both discovered the candidate paths and injected them — was split in two:

```
┌──────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────┐
│                Before (≤ v1.0.3)                 │                                      After (main)                                      │
├──────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SimpleCov.add_not_loaded_files(result) (private) │ (removed)                                                                              │
├──────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SimpleCov#discover_unloaded_paths (private)      │ SimpleCov::UnloadedFileInjector.discover(globs, root:, reject:) (public)               │
├──────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SimpleCov#inject_unloaded_files (private)        │ SimpleCov.inject_unloaded_files(coverage, paths, synthesize: nil, lines: nil) (public) │
└──────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────┘
```

Note that add_not_loaded_files was declared under private as far back as v0.22.0 — the contrib has always been reaching past the public API, which is why the removal wasn't caught by a version constraint or deprecation cycle. The replacement SimpleCov.inject_unloaded_files is deliberately public (it's the seam the merger injects through), so this is a stable target.

## Impact

This is worse than losing the coverage metric. extract_lines_pct powers a deprecated total-coverage tag, but the NoMethodError propagates all the way out through RSpec::Core::Runner#run_specs, so:

1. The RSpec process exits non-zero regardless of whether the specs passed.
2. SimpleCov's at_exit sees a pending exception and bails with Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected — no coverage report is written at all.
3. The failure is silent about its true origin; CI looks like a spec-suite failure.

## Suggested fix

Two independent changes, both worth making:

1. Support the new API (result_extractor.rb)
```ruby
def __dd_peek_result
  unless datadog_configuration[:enabled]
    Datadog.logger.debug("SimpleCov instrumentation is disabled")
    return nil
  end

  coverage = ::SimpleCov::UselessResultsRemover.call(
    ::SimpleCov::ResultAdapter.call(::Coverage.peek_result)
  )

  __dd_build_result(coverage)
end

private

def __dd_build_result(coverage)
  # simplecov > 1.0.3: injection is public and takes the paths explicitly.
  if ::SimpleCov.respond_to?(:inject_unloaded_files)
    coverage, not_loaded = ::SimpleCov.inject_unloaded_files(coverage, __dd_tracked_paths)
    return ::SimpleCov::Result.new(coverage, not_loaded_files: not_loaded)
  end

  # simplecov <= 1.0.3: private helper that also did the discovery.
  return ::SimpleCov::Result.new(coverage) unless ::SimpleCov.respond_to?(:add_not_loaded_files, true)

  processed = ::SimpleCov.send(:add_not_loaded_files, coverage)
  if processed.is_a?(::Array)
    ::SimpleCov::Result.new(processed[0], not_loaded_files: processed[1])
  else
    ::SimpleCov::Result.new(processed)
  end
end

def __dd_tracked_paths
  globs = [::SimpleCov.tracked_files, *::SimpleCov.cover_globs].compact
  return [] if globs.empty?

  ::SimpleCov::UnloadedFileInjector.discover(globs, root: ::SimpleCov.root)
end
```

tracked_files, cover_globs, and root are all public configuration readers, and UnloadedFileInjector.discover is public. SimpleCov::Result#initialize applies the configured filters itself, so anything discover over-collects is dropped at Result construction — no need to replicate SimpleCov's private reject: filtering.

2. Don't let a deprecated metric kill the test run (deprecated_total_coverage_metric.rb)

Whatever the fix above, this integration reaches into SimpleCov internals and should be assumed to break again on future SimpleCov releases. extract_lines_pct is best-effort telemetry for a deprecated tag; it should degrade to a debug log rather than escape into TestSession#finish:

```ruby
result = begin
  ::SimpleCov.__dd_peek_result
rescue StandardError => e
  Datadog.logger.debug("Failed to extract SimpleCov result: #{e.class}: #{e.message}")
  nil
end
```

An equivalent guard around the on_test_session_finished call site would work too.

Reproduction

# Gemfile
gem "datadog-ci", "~> 1.36"
gem "simplecov", github: "simplecov-ruby/simplecov", branch: "main"

Enable `Datadog.configure { |c| c.ci.enabled = true }` with the RSpec integration and run any spec file with SimpleCov started. Reverting simplecov to 1.0.3 makes it pass.

Related

- simplecov commit f315147 — "Inject unloaded files once at the merge point"
- simplecov issue #1250 — the parallel-run motivation for the move

anmarchenko:
## What changed

- Use SimpleCov's public unloaded-file injection API when available, while retaining the legacy path for SimpleCov 1.0.3 and older.
- Support both old and new `SimpleCov::Result` constructor shapes.
- Treat total-coverage extraction as best-effort: swallow extraction errors and emit a warning instead of aborting the test runner.
- Add regression coverage for the new API and failure isolation.
- Refresh all appraisal lockfiles, including SimpleCov 1.1.0 appraisals.

## Why

SimpleCov removed the private `add_not_loaded_files` helper after 1.0.3. The resulting `NoMethodError` escaped from test-session teardown, failed otherwise successful test processes, and prevented SimpleCov from producing its own report.

The integration now targets the replacement public API and keeps explicit compatibility with older SimpleCov releases. The deprecated total-coverage metric can no longer take down CI if SimpleCov changes again.

## Validation

- `bundle exec rake ci` on Ruby 3.4.5 with the refreshed appraisal set
- SimpleCov 0.22.0 appraisal: 23 examples, 0 failures
- SimpleCov 1.1.0 appraisal: 23 examples, 0 failures
- SimpleCov 1.0.3 exercised by the root CI bundle
- `bundle exec rake steep:check`
- StandardRB
- `git diff --check`

Closes #567

### DIFF
--- a/gemfiles/ruby_2.7_cucumber_3.gemfile.lock
+++ b/gemfiles/ruby_2.7_cucumber_3.gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DataDog/dd-trace-rb.git
-  revision: 1aa96bcbe81233e0a9a75399f34fd72102d59913
+  revision: 1541704c183d44a942d28af7302030aa3e30da71
   branch: master
   specs:
     datadog (2.41.0.dev)
@@ -65,7 +65,7 @@ GEM
     ffi (1.17.4)
     gherkin (5.1.0)
     hashdiff (1.2.1)
-    io-console (0.9.1)
+    io-console (0.9.2)
     irb (1.18.0)
       pp (>= 0.6.0)
       prism (>= 1.3.0)
--- a/gemfiles/ruby_2.7_cucumber_4.gemfile.lock
+++ b/gemfiles/ruby_2.7_cucumber_4.gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DataDog/dd-trace-rb.git
-  revision: 1aa96bcbe81233e0a9a75399f34fd72102d59913
+  revision: 1541704c183d44a942d28af7302030aa3e30da71
   branch: master
   specs:
     datadog (2.41.0.dev)
@@ -95,7 +95,7 @@ GEM
     hashdiff (1.2.1)
     i18n (1.14.8)
       concurrent-ruby (~> 1.0)
-    io-console (0.9.1)
+    io-console (0.9.2)
     irb (1.18.0)
       pp (>= 0.6.0)
       prism (>= 1.3.0)
--- a/gemfiles/ruby_2.7_cucumber_5.gemfile.lock
+++ b/gemfiles/ruby_2.7_cucumber_5.gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DataDog/dd-trace-rb.git
-  revision: 1aa96bcbe81233e0a9a75399f34fd72102d59913
+  revision: 1541704c183d44a942d28af7302030aa3e30da71
   branch: master
   specs:
     datadog (2.41.0.dev)
@@ -95,7 +95,7 @@ GEM
     hashdiff (1.2.1)
     i18n (1.14.8)
       concurrent-ruby (~> 1.0)
-    io-console (0.9.1)
+    io-console (0.9.2)
     irb (1.18.0)
       pp (>= 0.6.0)
       prism (>= 1.3.0)
--- a/gemfiles/ruby_2.7_cucumber_6.gemfile.lock
+++ b/gemfiles/ruby_2.7_cucumber_6.gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DataDog/dd-trace-rb.git
-  revision: 1aa96bcbe81233e0a9a75399f34fd72102d59913
+  revision: 1541704c183d44a942d28af7302030aa3e30da71
   branch: master
   specs:
     datadog (2.41.0.dev)
@@ -96,7 +96,7 @@ GEM
     hashdiff (1.2.1)
     i18n (1.14.8)
       concurrent-ruby (~> 1.0)
-    io-console (0.9.1)
+    io-console (0.9.2)
     irb (1.18.0)
       pp (>= 0.6.0)
       prism (>= 1.3.0)
--- a/gemfiles/ruby_2.7_cucumber_7.gemfile.lock
+++ b/gemfiles/ruby_2.7_cucumber_7.gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DataDog/dd-trace-rb.git
-  revision: 1aa96bcbe81233e0a9a75399f34fd72102d59913
+  revision: 1541704c183d44a942d28af7302030aa3e30da71
   branch: master
   specs:
     datadog (2.41.0.dev)
@@ -77,7 +77,7 @@ GEM
       cgi (>= 0.3.3)
     ffi (1.17.4)
     hashdiff (1.2.1)
-    io-console (0.9.1)
+    io-console (0.9.2)
     irb (1.18.0)
       pp (>= 0.6.0)
       prism (>= 1.3.0)
--- a/gemfiles/ruby_2.7_cucumber_8.gemfile.lock
+++ b/gemfiles/ruby_2.7_cucumber_8.gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DataDog/dd-trace-rb.git
-  revision: 1aa96bcbe81233e0a9a75399f34fd72102d59913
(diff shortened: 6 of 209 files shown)


anmarchenko:
## E2E Test Report: SUCCESS ✅

**Tested by:** Shepherd Agent (autonomous QA for Datadog Test Optimization)

### Test Environment

- **Method:** Local end-to-end testing with the Sidekiq Minitest playground, Ruby 3.3.5, and mockdog
- **Revision tested:** `2c8635cbec14065dbfa0562b5cce3dd44f2a7d50`
- **Dependency branch:** `anmarchenko/fix-simplecov-compatibility`

### Results

| Check | Status | Evidence |
| --- | --- | --- |
| SimpleCov 1.1.0 public unloaded-file injection path | ✅ | 423 runs, 1,427 assertions, 0 failures, 0 errors |
| SimpleCov 0.22.0 legacy compatibility path | ✅ | 423 runs, 1,427 assertions, 0 failures, 0 errors |
| Coverage extraction at session teardown | ✅ | Session metrics contained `test.code_coverage.lines_pct` (90.94% with 1.1.0; 91.02% with 0.22.0) |
| Former teardown regression | ✅ | No `add_not_loaded_files` `NoMethodError`, extraction warning, or test-runner abort occurred |
| Test Optimization intake | ✅ | Each run produced one passing session with 423 passing tests; mockdog received 579 test-cycle events |
| Datadog UI verification | N/A | Runs targeted local mockdog, so no Datadog UI data was expected; this PR does not change backend API requests or response handling |

### Issues Found

No product issues were found. The first attempt encountered an environment-only Redis connection failure because the cloned playground had not been fully set up. Running `crook setup sidekiq` started Redis; both intended verification runs then passed unchanged.

### Test Methodology

1. Injected the PR branch with Crook using `--dep datadog-ci-rb=anmarchenko/fix-simplecov-compatibility` and enabled debug logging.
2. Ran the full Sidekiq suite with its existing SimpleCov 0.22.0 dependency to verify the legacy path.
3. Temporarily resolved SimpleCov 1.1.0 and reran the full suite to exercise the public unloaded-file injection API that fixes the reported regression.
4. Inspected debug logs for coverage extraction, the former exception, and intake requests; then verified parsed mockdog session/test spans with `query-spans`.
5. Restored the playground dependency lockfile and confirmed both worktrees were clean.

---

_This E2E test was performed by Shepherd - autonomous QA agent for Datadog Test Optimization._

juan-fernandez (approved):

PR #615, opened 2026-09-11
add: dynamic retries for failed tests

gnufede:
**What does this PR do?**
Add dynamic Auto Test Retries (ATR) budgets based on test duration, instead of the flat per-test retry limit. When enabled, the number of retries allowed for a test is determined by the duration of its initial attempt, using the same duration buckets as Early Flake Detection (5s / 10s / 30s / 5m / >5m).

Two new env vars control the behavior:
- DD_CIVISIBILITY_DYNAMIC_ATR_ENABLED — enables duration-based ATR budgets. When unset/false, ATR stays on its existing flat-limit path.
- DD_CIVISIBILITY_DYNAMIC_ATR_BUCKETS — optionally overrides the five duration-based retry budgets with five positive comma-separated integers in [1, 20]. When unset/empty, the EFD retry settings from the backend are used.

The new handler classifies each test once (by its initial-attempt duration) and caches the resulting max-retries count for the lifetime of that test. The duration bucket index is computed via shared helpers on the EFD settings type, so bucket boundaries stay consistent with EFD (the existing EFD retry handler is refactored to use them too).

Telemetry: records a `dynamic_atr_retries.enabled` count metric with a `has_custom_buckets` tag when dynamic ATR is enabled.

Feature parity with dd-trace-py PR #20028.


lib/datadog/ci/test_retries/dynamic_atr_retries.rb
```diff
@@ -0,0 +1,49 @@
...
+      # buckets as Early Flake Detection (5s / 10s / 30s / 5m / >5m). The budgets can be
+      # overridden with DD_CIVISIBILITY_DYNAMIC_ATR_BUCKETS (five comma-separated integers
+      # in [1, 20]); when unset, the EFD slow-test-retry settings from the backend are used.
+      module DynamicATRRetries
+        BUCKET_COUNT = 5
+        MAX_RETRIES_PER_BUCKET = 20
+
+        # @return [Boolean] whether duration-based ATR retry budgets are enabled
+        def self.enabled?
+          Utils::Parsing.convert_to_bool(ENV[Ext::Settings::ENV_DYNAMIC_ATR_ENABLED])
```
anmarchenko: this should go via settings.rb and be injected into the TestRetries::Component

lib/datadog/ci/test_retries/dynamic_atr_retries.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+
+        # @return [Array<Integer>?, nil] configured ATR retry buckets, or nil to use the EFD retry settings
+        def self.buckets
+          raw = ENV[Ext::Settings::ENV_DYNAMIC_ATR_BUCKETS]
```
anmarchenko: this should go via settings.rb and be injected into the TestRetries::Component

lib/datadog/ci/test_retries/driver/retry_failed_dynamic.rb
```diff
@@ -0,0 +1,70 @@
...
+          def initialize(slow_test_retries, retries_buckets: nil)
+            @slow_test_retries = slow_test_retries
+            @retries_buckets = retries_buckets
+            @attempts = 0
+            @passed_once = false
+            @max_attempts = nil
+          end
+
+          def should_retry?
+            ensure_classified
```
anmarchenko: this is a method with side effect - not great, better to set the default state in the initializer

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## [Unreleased]
 
+### Added
```
anmarchenko: do not add CHANGELOG entries on commit - we compile them at release time

add this information to AGENTS.md

anmarchenko (approved):

